PR #91, opened 2021-06-04
Feat/sdk auth

Jakub-Vacek:
## Description

This PR enables SDK authentication using SDK token value from environment value

## Types of changes

## Checklist:

### DIFF
--- a/src/client/registry.ts
+++ b/src/client/registry.ts
@@ -98,7 +98,7 @@ export function loadSdkAuthToken(): string | undefined {
     return;
   }
   const token = loadedToken.trim();
-  const tokenRegexp = /^(sfs)_([^_]+)_([0-9A-F]{8})/i;
+  const tokenRegexp = /^(sfs)_([^_]+)_([0-9A-F]{8})$/i;
   if (!tokenRegexp.test(token)) {
     registryDebug(
       `Value in environment variable ${tokenEnvName} is not valid SDK authentization token`
--- a/src/client/registry.ts
+++ b/src/client/registry.ts
@@ -1,8 +1,11 @@
 import { MapDocumentNode } from '@superfaceai/ast';
+import createDebug from 'debug';
 import * as zod from 'zod';
 
 import { isProviderJson, ProviderJson } from '../internal';
 import { HttpClient } from '../lib/http';
+
+const registryDebug = createDebug('superface:Registry');
 
 export interface RegistryProviderInfo {
   url: string;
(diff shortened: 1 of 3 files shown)


src/client/registry.ts
```diff
@@ -1,9 +1,12 @@
 import { MapDocumentNode } from '@superfaceai/ast';
+import createDebug from 'debug';
 import * as zod from 'zod';
 
 import { isProviderJson, ProviderJson } from '../internal';
 import { HttpClient } from '../lib/http';
 
+const registryDebug = createDebug('superface:Registry');
```
lukas-valenta, line 8: i think all other are lowercase, let's keep it the same everywhere

src/client/registry.ts
```diff
@@ -85,6 +88,26 @@ const bindResponseValidator = zod.object({
...
+  const tokenEnvName = 'SUPERFACE_SDK_TOKEN';
+  //Load superface token
+  const loadedToken = process.env[tokenEnvName];
+  if (!loadedToken) {
+    registryDebug(`Environment variable ${tokenEnvName} not found`)
+
+    return;
+  }
+  const token = loadedToken.trim();
+  const tokenRegexp = /^(sfs)_(.+)_([0-9A-F]{8}$)/i;
```
TheEdward162, line 101: Greedy regex parser goes brrr. Pls replace `.` with `[^_]` at least.

TheEdward162: Also the `$` doesn't need to be in the last group

Jakub-Vacek: I just reused regex from BE 🤨🤷

src/client/registry.ts
Jakub-Vacek: At least on is upper cased https://github.com/superfaceai/one-sdk-js/blob/main/src/client/profile-provider.ts#L49
I will fix it to lower case everywhere :)

src/client/registry.ts
```diff
@@ -98,7 +98,7 @@ export function loadSdkAuthToken(): string | undefined {
     return;
   }
   const token = loadedToken.trim();
-  const tokenRegexp = /^(sfs)_(.+)_([0-9A-F]{8}$)/i;
+  const tokenRegexp = /^(sfs)_([^_]+)_([0-9A-F]{8})/i;
```
TheEdward162, line 101: I meant to keep the `$` but to move it out of the group :)

TheEdward162 (approved):

PR #92, opened 2021-06-09
feat: Very rough first draft of event system

lukas-valenta:
## Description
This is the event system that will be used to observe and intercept lifecycle events

## Motivation and Context

We need an event system to handle stuff like sending metrics or request retrying

## Types of changes

## Checklist:


TheEdward162 (commented):
This `on` hook doesn't really make sense to me. It's entire point is to allow retrying the same call with the same parameters, which I don't really see the point of.

In my opinion the hooks would make more sense if there were only pre- and post-hooks like this:

```ts
type EventContextBase = {
  readonly time: number
};

type BeforeHookResult<Target extends (...args: any) => any> = {
  kind: 'continue' // continues the execution, no change from this hook
} | {
  kind: 'modify', // modifies the arguments and continues the execution
  newArgs: Parameters<Target>
} | {
  kind: 'abort', // aborts the execution, no other hooks are run, control returns back to the intercepted function caller
  newResult: ReturnType<Target>
};

type BeforeHook<EventContext extends EventContextBase, Target extends (...args: any) => any> = (
  context: EventContext,
  args: Parameters<Target>
) => BeforeHookResult<Target>;

type AfterHookResult<Target extends (...args: any) => any> = {
  kind: 'continue' // no change
} | {
  kind: 'modify', // modifies the result of the target
  newResult: ReturnType<Target>
} | {
  kind: 'retry', // retries the call, possibly with new parameters
  newArgs?: Parameters<Target>
};

type AfterHook<EventContext extends EventContextBase, Target extends (...args: any) => any> = (
  context: EventContext,
  // the arguments the target was called with (after modification before hooks)
  args: Parameters<Target>,
  result: ReturnType<Target>
) => AfterHookResult<Target>;
```

then the circuit breaker would look something like this:

```ts
function hookCircuitBreaker() {
  const thisIsThePolice = new CircuitBreakerPolicy(/* ... */);

  events.on(
    'pre-fetch',
    { priority: 1 }, // I'd keep the object here in case we wanted to extend it in the future, like give the interceptor a name for logging purposes or something
    async (context, args) => {
      const resolution = thisIsThePolice.beforeExecution(
        {
          time: context.time,
          registryCacheAge: 0 // just imagine for now that this is available to the hook somehow
        }
      );

      let newArgs: undefined | typeof args;
      const resolveTimeout = (timeout: number) => {
        // setup timeout
        if (timeout > 0) {
          // important not to mutate the source arguments
          newArgs = clone(args);
          // or similar
          newArgs[1].timeout = timeout;
        }
      };

      if (resolution.kind === 'continue') {
        resolveTimeout(resolution.timeout);
      } else if (resolution.kind === 'backoff') {
        resolveTimeout(resolution.timeout);

        // sleep for the timeout
        await sleep(resolution.timeout);
      } else if (resolution.kind === 'recache') {
        resolveTimeout(resolution.timeout);

        await recacheSomehow(resolution.newRegistry);
      } else if (resolution.kind === 'abort') {
        return {
          kind: 'abort',
          newResult: Promise.reject(resolution.reason)
        };
      } else if (resolution.kind === 'switch-provider') {
        // idk yet
      }

      if (newArgs !== undefined) {
        return {
          kind: 'modify',
          newArgs
        };
      } else {
        return {
          kind: 'continue'
        };
      }
    }
  );

  events.on(
    'post-fetch',
    { priority: 1 },
    async (context, _args, result) => {
      return result.match(
        _result => {
          const resolution = thisIsThePolice.afterSuccess(
            {
              time: context.time,
              registryCacheAge: 0
            }
          );

          if (resolution.kind === 'continue') {
            return {
              kind: 'continue'
            };
          }
        },
        _error => {
          const resolution = thisIsThePolice.afterFailure(
            // choose based on the `error`
            {
              time: context.time,
              registryCacheAge: 0,
              kind: 'network',
              issue: 'timeout'
            }
          );

          if (resolution.kind === 'retry') {
            return {
              kind: 'retry'
            }
          } else if (resolution.kind === 'abort') {
            return {
              kind: 'modify',
              newResult: Promise.reject(resolution.reason)
            }
          } else if (resolution.kind === 'continue') {
            return {
              kind: 'continue'
            }
          }
        }
      );
    }
  );
}

```

src/lib/events.ts
```diff
@@ -0,0 +1,204 @@
...
+  fetch: (args: ReturnType<FetchInstance['fetch']>) => boolean;
+  'post-fetch': (
+    args: ReturnType<FetchInstance['fetch']>
+  ) => ReturnType<FetchInstance['fetch']>;
+};
+
+type EventListeners = {
+  [E in keyof EventParams]?: PriorityCallbackTuple[];
+};
+type Filter = { usecase?: string; profile?: string };
```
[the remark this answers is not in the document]
TheEdward162: See the main review comment, specifically the `EventContext` parameter which we would extend with the event specific information, like usecase and profile.

Then the hooks would filter based on that.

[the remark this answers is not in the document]
TheEdward162: Specifically the failure policy hooks would look for the policies in some place they would capture in their closures and decide based on the existence of the policies on what to do.

src/client/profile-provider.ts
```diff
@@ -97,6 +101,10 @@ export class BoundProfileProvider {
     input?: TInput
   ): Promise<Result<TResult, ProfileParameterError | MapInterpreterError>> {
     // compose and validate the input
+    Reflect.set(this.fetchInstance, 'metadata', {
+      profile: profileAstId(this.profileAst),
+      usecase,
+    });
```
TheEdward162, line 107: Why is this needed?

src/lib/events.ts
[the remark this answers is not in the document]
lukas-valenta: you can do both and filtering by usecase will be used often, so i thought to make it first class

src/client/profile-provider.ts
lukas-valenta: to let fetch interceptors know which usecase and profile you're calling

lukas-valenta:
you're probably right on the hook resolution and the 'on' hook

src/lib/events.ts
```diff
@@ -60,134 +129,149 @@ class Events {
...
+      if (hookResult.kind === 'abort') {
+        return hookResult.newResult;
+      }
+
+      if (hookResult.kind === 'continue') {
+        // DO NOTHING YAY!
+      }
+    }
+
+    let result = (await originalFunction.apply(this, args)) as ReturnType<
```
TheEdward162, line 203: Shouldn't this take `functionArgs`?

src/lib/events.ts
```diff
@@ -60,134 +129,149 @@ class Events {
...
+          return hookResult.newResult;
+        }
 
-export function eventInterceptor(
-  eventMetadata: EventMetadata
+        if (hookResult.kind === 'retry') {
+          if (hookResult.newArgs !== undefined) {
+            result = await originalFunction?.apply(this, hookResult.newArgs);
+          } else {
+            result = await originalFunction?.apply(this, args);
```
TheEdward162, line 235: Also `functionArgs`?

TheEdward162 (approved):

TheEdward162:
Eh, I didn't want to approve, just comment

src/lib/events.ts
lukas-valenta: indeed

PR #95, opened 2021-06-14
feat/priority array basic setup

Jakub-Vacek:
## Description

This PR adds provider priority array to super.json

## Types of changes

## Checklist:

### DIFF
--- a/src/internal/superjson/mutate.ts
+++ b/src/internal/superjson/mutate.ts
@@ -1,3 +1,4 @@
+import { err, ok, Result } from '../../lib';
 import { isEmptyRecord } from '../interpreter/variables';
 import {
   normalizeProfileSettings,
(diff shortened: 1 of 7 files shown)


src/internal/superjson/mutate.ts
```diff
@@ -303,3 +303,28 @@ export function mergeSecurity(
 
   return result;
 }
+
+//TODO: should we check if target profile and providers property has configured providers from priority array?
+export function addPriority(
```
TheEdward162, line 308: I think we should definitely check if the provider is configured in `providers` section of super.json.

Checking for it being configured in profile settings is I think reasonable too, but overall I'm not sure.

Also I think this should return `false` when the previous array is the same as the new one (e.g. check the element equality, including the order).

And finally, I'm not entirely sure if creating the profile entry (with version 0.0.0) is the right move. I know it's done in profileProvider mutator above, but I'm not sure about that one either. Maybe these should return `Result<boolean, E>` and error on profile not existing or something 🤷‍♀️ (and cache would be invalidated only when they return `ok(true)`.

Jakub-Vacek: @TheEdward162 I added some checks :)

lukas-valenta (approved):

PR #96, opened 2021-06-15
feat: Rework config hash and add anonymization

lukas-valenta:
## Description

Creates a function to get a subset of super.json to send with metrics. Reworks config hash function to work with this subset.

## Motivation and Context

We need to send the subset to backend with metrics.

## Types of changes

## Checklist:

### DIFF
--- a/src/internal/superjson/schema.ts
+++ b/src/internal/superjson/schema.ts
@@ -290,3 +290,18 @@ export type NormalizedProfileProviderSettings = zod.infer<
 export type NormalizedProviderSettings = zod.infer<
   typeof normalizedProviderSettings
 >;
+
+export type AnonymizedSuperJsonDocument = {
+  profiles: Record<
+    string,
+    {
+      version: string | 'file';
+      providers: {
+        provider: string;
+        priority: number;
+        version?: string | 'file';
+      }[];
+    }
+  >;
+  providers: string[];
+};
--- a/src/internal/superjson/superjson.ts
+++ b/src/internal/superjson/superjson.ts
@@ -19,6 +19,7 @@ import {
 } from './mutate';
 import { normalizeSuperJsonDocument } from './normalize';
 import {
+  AnonymizedSuperJsonDocument,
   NormalizedSuperJsonDocument,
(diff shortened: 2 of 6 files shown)


tsconfig.json
```diff
@@ -11,7 +11,7 @@
     "noFallthroughCasesInSwitch": true,
     "noImplicitAny": true,
     "noImplicitReturns": true,
-    "noImplicitOverride": true,
+    "noImplicitOverride": false,
```
Jakub-Vacek, line 14: 😂

lukas-valenta: only until the prettier fix releases!

src/internal/superjson/schema.ts
```diff
@@ -290,3 +290,18 @@ export type NormalizedProfileProviderSettings = zod.infer<
...
+
+export type AnonymizedSuperJsonDocument = {
+  profiles: Record<
+    string,
+    {
+      version: string | 'file';
+      providers: {
+        provider: string;
+        priority: number;
+        version?: string | 'file';
```
Jakub-Vacek, line 302: This is version of map?

lukas-valenta: i think so

src/internal/superjson/superjson.ts
```diff
@@ -256,44 +257,74 @@ export class SuperJson {
...
-          if ('file' in info) {
-            return `${provider}:${info.file}`;
-          } else {
-            return `${provider}:${info.mapVariant ?? ''}-${
-              info.mapRevision ?? ''
-            }`;
-          }
+    return {
+      profiles,
+      providers: Object.keys(this.normalized.providers),
```
Jakub-Vacek, line 294: Can normalized providers contain secrets - loaded from env or directly stored in super.json

Jakub-Vacek (approved):

src/internal/superjson/superjson.ts
lukas-valenta: no, this is just keys == names

PR #99, opened 2021-06-21
Feature/integrate events failure

Jakub-Vacek:
## Types of changes

## Checklist:

- [ ] I have read the **CONTRIBUTION_GUIDE** document.


src/client/failure/event-adapter.test.ts
```diff
@@ -1,11 +1,18 @@
 import { MapDocumentNode, ProfileDocumentNode } from '@superfaceai/ast';
+import {
+  ProfileConfiguration,
+  Provider,
+  ProviderConfiguration,
+  UseCase,
+} from '@superfaceai/one-sdk';
+import { SuperfaceClientBase } from '@superfaceai/one-sdk/dist/client/public/client';
```
lukas-valenta, line 8: these are not the imports you're looking for :grin:

Jakub-Vacek: Yep, this file is a mess. I am rewriting it now 🙂

src/client/client.ts
```diff
@@ -27,17 +31,14 @@ export abstract class SuperfaceClientBase {
 
     if (SUPER_CACHE[superCacheKey] === undefined) {
       SUPER_CACHE[superCacheKey] = SuperJson.loadSync(superCacheKey).unwrap();
+    } else {
+      //TODO: better way of cummunicating this to user.
+      console.warn('Multiple SuperfaceClient bad');
```
TheEdward162, line 36: 🤣

src/client/client.ts
```diff
@@ -133,8 +134,128 @@ export abstract class SuperfaceClientBase {
...
+      { providersOfUsecase: Record<string, FailurePolicy> }
+    > = {};
+    let policy: FailurePolicy;
+    for (const [profile, profileSettings] of Object.entries(
+      this.superJson.normalized.profiles
+    )) {
+      //Set failoverPolicy
+      const priority = profileSettings.priority;
+      if (!priority.every(p => this.superJson.normalized.providers[p])) {
+        throw new SDKExecutionError(
```
TheEdward162, line 176: Duplicate code :/

src/client/usecase.ts
```diff
@@ -70,20 +76,20 @@ export class UseCase extends UseCaseBase {
     const boundProfileProvider = await this.bind(options);
 
     // TOOD: rewrap the errors for public consumption?
-    return boundProfileProvider.perform<TInput, TOutput>(this.name, input);
+    return await boundProfileProvider.perform<TInput, TOutput>(this.name, input);
```
TheEdward162, line 79: I'm afraid to ask but.. what breaks without this await?

Jakub-Vacek: Nope, I was just trying to localize the concurrency bug :)

src/client/failure/policies.ts
```diff
@@ -12,6 +12,71 @@ import {
   SuccessResolution,
 } from './resolution';
 
+export class Router {
+  constructor(
+    private readonly providersOfUseCase: Record<string, FailurePolicy>,
+    private readonly priority: string[],
+    private currentProvider?: string
```
TheEdward162, line 19: Why is this optional?

src/client/failure/policies.ts
```diff
@@ -12,6 +12,71 @@ import {
...
+  public afterSuccess(info: ExecutionSuccess): SuccessResolution {
+    if (!this.currentProvider) {
+      throw 'Property currentProvider is not set in Router instance';
+    }
+
+    return this.providersOfUseCase[this.currentProvider].afterSuccess(info);
+  }
+
+  //TODO: Should we handle provider reset here?
+  public reset(): void {}
```
TheEdward162, line 77: Imo this should reset itself to the first provider and reset all inner policies

src/client/failure/policies.ts
```diff
@@ -12,6 +12,71 @@ import {
   SuccessResolution,
 } from './resolution';
 
+export class Router {
```
TheEdward162, line 15: Can we please name this something more descriptive? Like `FailurePolicyRouter`.

src/client/failure/policies.ts
```diff
@@ -12,6 +12,71 @@ import {
...
+    // TODO: check if exists
+    this.currentProvider = provider;
+  }
+
+  public beforeExecution(info: ExecutionInfo): ExecutionResolution {
+    if (!this.currentProvider) {
+      throw 'Property currentProvider is not set in Router instance';
+    }
+    const innerResolution =
+      this.providersOfUseCase[this.currentProvider].beforeExecution(info);
```
TheEdward162, line 36: At some point we should also attempt going back to previous providers. I think an iteration of the return values of `beforeExecution` here should do the trick.

Jakub-Vacek: @TheEdward162 I added something very similar and also test for it.

PR #100, opened 2021-07-02
Feature/metric reporting

lukas-valenta:
## Description

Adds metric reporting for Initialization, Perform and Failover

## Motivation and Context

We want metric reporting to show on dashboard and sending notifications

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [X] I have read the **CONTRIBUTION_GUIDE** document.
- [X] I haven't repeated the code. (DRY)


src/client/client.test.ts
```diff
@@ -98,34 +96,32 @@ describe('superface client', () => {
   it('caches super.json files correctly', () => {
     const client = new SuperfaceClient();
     expect(client.superJson.document).toEqual(MOCK_SUPERJSON);
-    expect(statSyncMock).toHaveBeenCalledTimes(1);
```
Jakub-Vacek, line 101: Isn't this like saying: const x = 3, expect(x === x).toBe(true)? 🤷

src/client/failure/event-adapter.ts
```diff
@@ -150,6 +158,15 @@ export function registerHooks(hookContext: HooksContext, events: Events): void {
         return { kind: 'retry' };
 
       case 'abort':
+        void events.emit('provider-switch', [
```
Jakub-Vacek, line 161: Are we sure that provider switch is happening? Can be this abort from AbortPolicy?

src/config.ts
```diff
@@ -0,0 +1,90 @@
+import createDebug from 'debug';
```
Jakub-Vacek, line 1: I like the centralized config approach 👍👍

src/client/client.test.ts
lukas-valenta: https://applifting.slack.com/archives/C01PVQBACB1/p1624269231008500

src/lib/reporter.test.ts
```diff
@@ -0,0 +1,147 @@
...
+      {
+        security: [],
+      },
+      client
+    );
+    const profile = await client.getProfile('test-profile');
+    await profile.getUseCase('Test').perform();
+    await profile.getUseCase('Test').perform();
+    expect(client).not.toBe(undefined);
+    await new Promise(resolve => setTimeout(resolve, 20000));
```
Jakub-Vacek, line 145: What we are testing here?

src/client/failure/event-adapter.ts
lukas-valenta: it's not happening. it's actually reporting that it can't switch, because there's not another provider configured. i suppose i should name the event better :thinking:

Jakub-Vacek: Name confused me. On the other hand - if the only difference is the name 🤷

src/lib/reporter.test.ts
lukas-valenta: nothing, there are no tests yet

src/client/client.test.ts
Jakub-Vacek: ✅

src/lib/reporter.test.ts
Jakub-Vacek: Well, that was funny to think about what do you want to do there😅

src/lib/reporter.ts
```diff
@@ -0,0 +1,308 @@
...
+    }
+    if (timeHasElapsed) {
+      return;
+    }
+    if (this.timer !== undefined) {
+      clearTimeout(this.timer);
+    }
+    this.timer = setTimeout(() => {
+      this.flush();
+    }, 1000);
```
TheEdward162, line 186: Shouldn't this be also configurable?

TheEdward162 (approved):

src/lib/reporter.test.ts
```diff
@@ -107,41 +145,524 @@ const mockMapDocument: MapDocumentNode = {
...
+  });
+
+  it('should report SDK Init', async () => {
     const client = new SuperfaceClient();
-    (client as any).boundCache[
-      '{"id":"test-profile","version":"1.0.0"}{"name":"testprovider","security":[]}'
-    ] = new BoundProfileProvider(
+    void client;
+    while (await eventEndpoint.isPending()) {
+      await new Promise(setImmediate);
```
TheEdward162, line 233: How is this different from `Promise.resolve()`?

TheEdward162: Overall, what's the reason behind these loops?

lukas-valenta: the main difference is that `Promise.resolve()` doesn't work for some reason, it will just loop endlessly until the heat death of the universe. the reason behind the loops is that the reporter calls are void promises, running somewhere in the background with no way to wait for them. if i did just loop, the event loop wouldn't execute, the setImmediate promise await gives it control back for a bit (to resolve any pending calls etc.). that's what i came up with anyway, if you have a better solution tell me

TheEdward162: Does `Promise.resolve().then(() => undefined)` do something different? The MDN wiki isn't exactly clear. My main issue is that the `setImmediate` api is NodeJS only and it's another hurdle we'd have to get through if we ever get to that sweet browser integration.

PR #105, opened 2021-07-08
chore: pass request error thru policy

Jakub-Vacek:
## Description

This PR adds formatFailureReason function to enable passing error thru Retry and Abort policy

## Types of changes

## Checklist:

### DIFF
--- a/src/client/failure/policies.ts
+++ b/src/client/failure/policies.ts
@@ -168,9 +168,9 @@ export class AbortPolicy extends FailurePolicy {
     return { kind: 'continue', timeout: 30_000 };
   }
 
-  override afterFailure(_info: ExecutionFailure): FailureResolution {
+  override afterFailure(info: ExecutionFailure): FailureResolution {
     //TODO: Eda said this maybe should be continue
-    return { kind: 'abort', reason: 'abort policy selected' };
+    return { kind: 'abort', reason: this.formatFailureReason(info) };
   }
 
   override afterSuccess(_info: ExecutionSuccess): SuccessResolution {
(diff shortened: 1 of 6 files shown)


src/client/failure/policies.ts
```diff
@@ -325,7 +334,12 @@ export class CircuitBreakerPolicy extends FailurePolicy {
     if (this.state === 'half-open') {
       this.open(info.time);
 
-      return { kind: 'abort', reason: 'circuit breaker is open' };
+      return {
+        kind: 'abort',
+        reason: `Circuit breaker is open. Last error from provider: ${this.formatFailureReason(
+          info
+        )}`,
```
TheEdward162, line 341: So do we need both `formatFailureReason` and passing the original error in `FailurePolicyError`?

Jakub-Vacek: I think we do, maybe we could do reason formatting in FailurePolicyError.
In some (pre-fetch and open circuit-breaker) cases we don't have original error so reason is all we have 🤷

src/client/failure/policy.ts
```diff
@@ -1,3 +1,6 @@
+import { HttpResponse } from '@superfaceai/one-sdk';
```
lukas-valenta, line 1: fix this pls and do something about this, this happens to you a lot for some reason :suspect:

Jakub-Vacek: Yep, my vscode is borked when it comes to imports, already tried reinstall. Good thing is that lint in CI catches this stupid stuff.

src/client/failure/policies.ts
TheEdward162: `pre-fetch` doesn't have an original error. If you are doing abort in pre-fetch it's because you are doing failover.

Open circuit-breaker either has an error that opened it right there, or it was opened last time and the shadowed by a failover. In the first case it's trivial, in the second that error wouldn't be available anyway.

Jakub-Vacek: Yes, and that's why originalError is optional in FailurePolicyError and we need reason 🤷

TheEdward162: Yes but do we need to include the error in the reason when the error (if any) will be part of the `FailurePolicyError`?

lukas-valenta (approved):

PR #107, opened 2021-07-09
feat: Add toString() methods to errors

martinalbert:
## Description

Add `toString()` methods to error classes that don't contain this method.

## Motivation and Context

Error formatting.

## Types of changes

## Checklist:

### DIFF
--- a/src/internal/interpreter/map-interpreter.errors.ts
+++ b/src/internal/interpreter/map-interpreter.errors.ts
@@ -108,11 +108,6 @@ export class HTTPError extends MapInterpreterErrorBase {
         ? `Original Map Location: Line ${this.metadata.node.location.line}, column ${this.metadata.node.location.column}`
         : undefined,
       this.request?.url ? `Request URL: ${this.request.url}` : undefined,
-      this.response?.headers
-        ? `Response headers:\n${Object.entries(this.response.headers)
-            .map(([key, value]) => `\t${key}: ${value}`)
-            .join('\n')}`
-        : undefined,
     ]
       .filter(line => !!line)
       .join('\n');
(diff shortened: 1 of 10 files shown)


src/internal/interpreter/map-interpreter.errors.ts
```diff
@@ -99,6 +99,29 @@ export class HTTPError extends MapInterpreterErrorBase {
...
+  override toString(): string {
+    return [
+      `${this.kind}: ${this.message}`,
+      this.astPath ? `AST Path: ${this.astPath.join('.')}` : undefined,
+      this.metadata?.node?.location
+        ? `Original Map Location: Line ${this.metadata.node.location.line}, column ${this.metadata.node.location.column}`
+        : undefined,
+      this.request?.url ? `Request URL: ${this.request.url}` : undefined,
+      this.request?.headers
+        ? `Request headers:\n${Object.entries(this.request.headers)
```
lukas-valenta, line 112: i wouldn't put this here, headers can contain sensitive information

martinalbert: make sense 👍🏻

martinalbert: What about response headers though?

lukas-valenta: same. there's sensitive debug for this

martinalbert: ok thanks, removed 🙌🏻

lukas-valenta (approved):

src/internal/interpreter/map-interpreter.errors.ts
TheEdward162: Doesn't the url also contain sensitive info? Or is this the url before query parameters would be resolved?

lukas-valenta: good point

TheEdward162 (approved):

PR #113, opened 2021-07-15
fix: profile provider mutate

Jakub-Vacek:
## Description

This PR fixes problems with priority and defaults resolution when adding new profile provider

## Types of changes

## Checklist:

### DIFF
--- a/src/internal/superjson/mutate.ts
+++ b/src/internal/superjson/mutate.ts
@@ -194,7 +194,19 @@ export function addProfile(
 
   return true;
 }
+function resolvePriorityAddition(
+  existingPriority: string[] | undefined,
+  newProvider: string
+): string[] {
+  if (!existingPriority || existingPriority.length === 0) {
+    return [newProvider];
+  }
+  if (!existingPriority.includes(newProvider)) {
+    return [...existingPriority, newProvider];
+  }
 
+  return existingPriority;
+}
 export function addProfileProvider(
   document: SuperJsonDocument,
   profileName: string,
@@ -220,6 +232,10 @@ export function addProfileProvider(
     targetedProfile.providers = {
       [providerName]: payload,
     };
+    targetedProfile.priority = resolvePriorityAddition(
+      targetedProfile.priority,
+      providerName
+    );
 
     return true;
   }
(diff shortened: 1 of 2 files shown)


src/internal/superjson/mutate.ts
```diff
@@ -257,32 +277,60 @@ export function addProfileProvider(
...
   if (typeof profileProvider === 'string') {
+    //Change
     defaults = payload.defaults;
-  } else if (profileProvider.defaults) {
-    if (!payload.defaults) {
-      defaults = targetedProfile.defaults;
-    } else {
+  } else {
+    if (profileProvider.defaults && payload.defaults) {
+      //Chanage
```
TheEdward162, line 295: Chanage lol

src/internal/superjson/mutate.ts
```diff
@@ -220,6 +232,10 @@ export function addProfileProvider(
     targetedProfile.providers = {
       [providerName]: payload,
     };
+    targetedProfile.priority = resolvePriorityAddition(
+      targetedProfile.priority,
+      providerName
+    );
```
TheEdward162, line 238: Is this needed in this branch as we update the targetedProfile above?

lukas-valenta (approved):

src/internal/superjson/mutate.ts
```diff
@@ -257,32 +277,60 @@ export function addProfileProvider(
...
   }
+  // if there no other keys and we changed defaults
+  if (
+    !('file' in payload) &&
+    !('mapVariant' in payload) &&
+    !('mapRevision' in payload) &&
+    defaults
+  ) {
+    targetedProfile.providers[providerName] = {
+      ...(typeof profileProvider === 'string' ? {} : profileProvider),
```
TheEdward162, line 313: Don't we need to preserve it as a `file` entry?

Jakub-Vacek: You mean like:
if typeof profileProvider === 'string' {
targetedProfile.providers[providerName] = {
file: profileProvider
defaults: payload.defaults
}
}

Jakub-Vacek: `    targetedProfile.providers[providerName] = {
      ...(typeof profileProvider === 'string' ? { file: profileProvider } : profileProvider),
      defaults,
    };`

TheEdward162: Yeah, otherwise I think we'd lose the path

TheEdward162 (approved):

PR #116, opened 2021-07-21
feat: Use Error subclasses

lukas-valenta:
## Description
Make every `throw` use an Error subclass

## Motivation and Context

We want Errors to be more descriptive, before this some errors threw plain Error or even just string

## Types of changes

## Checklist:

### DIFF
--- a/src/internal/errors.helpers.ts
+++ b/src/internal/errors.helpers.ts
@@ -1,0 +1,303 @@
+import { ZodError } from 'zod';
+
+import { SDKExecutionError } from './errors';
+import { BackoffKind, SecurityValues } from './superjson';
+
+export function assertError(error: unknown): Error {
+  if (typeof error === 'string') {
+    return new Error(error);
+  } else if (error instanceof Error) {
+    return error;
+  }
+
+  return new Error(JSON.stringify(error));
+}
+
+export function isZodError(error: Error): error is ZodError {
+  return error instanceof ZodError;
+}
+
+export function superJsonNotFoundError(
+  path: string,
+  error: Error
+): SDKExecutionError {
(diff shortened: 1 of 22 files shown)


src/internal/errors.helpers.ts
```diff
@@ -0,0 +1,327 @@
+import { ZodError } from 'zod';
+
+import { SDKExecutionError } from './errors';
+import { BackoffKind, SecurityValues } from './superjson';
+
+export function assertError(error: unknown): Error {
```
TheEdward162, line 6: I'd call this `wrapError` or `ensureErrorSubclass` or such.

src/internal/superjson/mutate.ts
```diff
@@ -515,11 +506,8 @@ export function addPriority(
       (value: string, index: number) => value === existingPriority[index]
     )
   ) {
-    return err(
-      new Error(
-        `Unable to set priority on profile "${profileName}" - existing priority is same as new priority`
-      )
-    );
+    // TODO: Shouldn't this return ok(false)?
+    return err(priorityIsTheSameError(profileName));
```
TheEdward162, line 510: I think it should, yes

src/internal/errors.helpers.ts
TheEdward162:
```suggestion
export function ensureErrorSubclass(error: unknown): Error {
```

src/internal/superjson/mutate.ts
TheEdward162:
```suggestion
    return ok(false);
```

TheEdward162 (approved):

PR #118, opened 2021-08-03
WIP: feat: Add map and profile parsing

lukas-valenta:
## Description

Adds map and profile on-the-fly parsing

## Motivation and Context

Makes it easier to author maps and profiles

## Types of changes

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
   },
   "dependencies": {
     "@superfaceai/ast": "0.0.23",
+    "@superfaceai/parser": "^0.0.19",
     "@typescript-eslint/eslint-plugin": "^4.27.0",
     "@typescript-eslint/parser": "^4.27.0",
     "abort-controller": "^3.0.0",
--- a/src/config.ts
+++ b/src/config.ts
@@ -24,6 +24,11 @@ export const DEFAULT_METRIC_DEBOUNCE_TIME = {
   min: 1000,
   max: 60000,
 };
+export const DEFAULT_CACHE_PATH = joinPath(
+  process.cwd(),
+  '.cache',
+  'superface'
+);
 
 // Extraction functions
 function getSuperfaceApiUrl(): string {
@@ -81,6 +86,7 @@ export type Config = {
   metricDebounceTimeMin: number;
   metricDebounceTimeMax: number;
   disableReporting: boolean;
+  cachePath: string;
 };
 
 let configCache: Config | undefined;
@@ -96,6 +102,7 @@ export const Config = (): Config => {
         process.env.NODE_ENV === 'test'
           ? true
           : !!process.env[DISABLE_REPORTING],
+      cachePath: DEFAULT_CACHE_PATH,
     };
   }
 
--- a/src/internal/parser.ts
+++ b/src/internal/parser.ts
@@ -1,0 +1,156 @@
+import {
+  assertMapDocumentNode,
+  assertProfileDocumentNode,
+  EXTENSIONS,
+  MapDocumentNode,
+  ProfileDocumentNode,
+} from '@superfaceai/ast';
+import { parseMap, parseProfile, Source } from '@superfaceai/parser';
+import { createHash } from 'crypto';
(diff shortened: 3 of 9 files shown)


src/config.ts
```diff
@@ -24,6 +24,11 @@ export const DEFAULT_METRIC_DEBOUNCE_TIME = {
   min: 1000,
   max: 60000,
 };
+export const DEFAULT_CACHE_PATH = joinPath(
+  process.cwd(),
+  '.cache',
+  'superface'
+);
```
Jakub-Vacek, line 31: Would creating .cache under superface dir make sense? We would have everything under same folder 🤷

lukas-valenta: i suppose. we talked about it, but i don't remember what we settled on. in that case we should update init to put a .gitignore in superface directory

Jakub-Vacek: I think we just listed options 🤷😅 
Agree on .gitignore

src/internal/parser.ts
```diff
@@ -0,0 +1,156 @@
...
+import { promises as fsp } from 'fs';
+import { join as joinPath } from 'path';
+
+import { Config } from '../config';
+
+export class Parser {
+  private static mapCache: Record<string, MapDocumentNode> = {};
+  private static profileCache: Record<string, ProfileDocumentNode> = {};
+
+  static async parseMap(
```
Jakub-Vacek, line 19: Could we have some sort of flag which would force parsing of the map (profile)? In CLI check/lint/test it would be cool to actually compile on every run to ensure correct results 🤷

src/internal/parser.ts
```diff
@@ -0,0 +1,156 @@
...
+
+    // If not, delete old parsed maps
+    const cachedFileRegex = new RegExp(
+      `${info.providerName}-[0-9a-f]+${EXTENSIONS.map.build}`
+    );
+    try {
+      for (const file of (await fsp.readdir(cachePath)).filter(cachedFile =>
+        cachedFileRegex.test(cachedFile)
+      )) {
+        await fsp.unlink(file);
```
Jakub-Vacek, line 65: Probably `unlink(joinPath(cachePath, file))` 🤷

src/internal/parser.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  static async parseMap(
+    input: string,
+    fileName: string,
+    info: {
+      profileName: string;
+      providerName: string;
+      scope?: string;
+    }
+  ): Promise<MapDocumentNode> {
```
TheEdward162, line 27: I don't quite understand why this requires the source `input` - ideally it should be able to cache based on the documentId so that you don't have to download the source when you want to check if a map is in the cache.

src/internal/parser.ts
lukas-valenta: yup, already fixed

src/internal/parser.ts
lukas-valenta: because it caches based on content, for local development

src/internal/parser.ts
lukas-valenta: sure

Jakub-Vacek: After looking at it more I think that my idea is stupid 😅 - if we hash the source we will recompile on every source change anyway. 🤷

lukas-valenta: true

package.json
```diff
@@ -47,7 +47,10 @@
     "typescript": "4.3.2"
   },
   "dependencies": {
-    "@superfaceai/ast": "0.0.23",
+    "@superfaceai/ast": "0.0.28",
+    "@superfaceai/parser": "^0.0.19",
+    "@typescript-eslint/eslint-plugin": "^4.27.0",
```
Jakub-Vacek, line 52: Didn't we removed this ESLint deps in some other PR?

lukas-valenta: looks like it's back :trollface: 
i'll fix it

Jakub-Vacek (approved):

PR #120, opened 2021-08-13
feat: superjson swap variant methods

TheEdward162:
## Description
Adds swap variant methods to super json for swapping provider and profile provider variants between local and remote while preserving other keys.

Renames existing add methods to merge as it better describes the functionality they provide - their behaviour does not change.

## Types of changes

## Checklist:


src/internal/superjson/mutate.ts
```diff
@@ -427,6 +542,50 @@ export function addProvider(
   return true;
 }
 
+export function swapProviderVariant(
+  document: SuperJsonDocument,
+  providerName: string,
+  variant: { kind: 'local'; file: string } | { kind: 'remote' }
+): boolean {
+  if (document.providers === undefined) {
+    document.providers = {};
```
Jakub-Vacek, line 551: Do we count this as "change"? 🤷

Jakub-Vacek (approved):

src/internal/superjson/mutate.ts
TheEdward162: I went with a "no" because it doesn't change the semantic view of the document, only changes `undefined` for `{}` which is semantically equal.

PR #121, opened 2021-08-17
feat: fetch provider info when map is local

TheEdward162:
## Description
It is now implemented to fetch provider info when a map is local.

## Motivation and Context
Previously it wasn't implemented.

## Types of changes

## Checklist:

- [ ] I have read the **CONTRIBUTION_GUIDE** document.

### DIFF
--- a/src/client/profile-provider.ts
+++ b/src/client/profile-provider.ts
@@ -9,7 +9,6 @@ import {
   SecurityScheme,
   SecurityType,
 } from '../internal';
-import { UnexpectedError } from '../internal/errors';
 import {
   invalidProfileError,
   invalidSecurityValuesError,
@@ -47,7 +46,7 @@ import { CrossFetch } from '../lib/fetch';
 import { MapInterpreterEventAdapter } from './failure/map-interpreter-adapter';
 import { ProfileConfiguration } from './profile';
 import { ProviderConfiguration } from './provider';
-import { fetchBind } from './registry';
+import { fetchBind, fetchProviderInfo } from './registry';
 
 function forceCast<T>(_: unknown): asserts _ is T {}
 
@@ -70,7 +69,7 @@ export class BoundProfileProvider {
     private readonly mapAst: MapDocumentNode,
     private readonly providerName: string,
     private readonly configuration: {
-      baseUrl?: string;
+      baseUrl: string;
       profileProviderSettings?: NormalizedProfileProviderSettings;
       security: SecurityConfiguration[];
     },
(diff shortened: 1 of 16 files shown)


src/client/registry.ts
```diff
@@ -55,17 +55,17 @@ export function assertIsRegistryProviderInfo(
...
 
-export async function fetchMapAST(url: string): Promise<MapDocumentNode> {
-  const fetchInstance = new CrossFetch();
-  const http = new HttpClient(fetchInstance);
-  registryDebug('Fetching Map AST from registry');
-  const { body } = await http.request(url, {
-    method: 'GET',
-    accept: 'application/json',
-  });
+// export async function fetchMapAST(url: string): Promise<MapDocumentNode> {
```
Jakub-Vacek, line 58: Do we want to keep it? 🤷

src/client/provider.ts
```diff
@@ -23,12 +22,10 @@ export class Provider {
 
   async configure(configuration: {
     security?: SecurityValues[];
-    serviceId?: string;
```
Jakub-Vacek, line 26: Removed because we don't use it or?🤷

src/client/profile-provider.ts
```diff
@@ -70,7 +69,7 @@ export class BoundProfileProvider {
     private readonly mapAst: MapDocumentNode,
     private readonly providerName: string,
     private readonly configuration: {
-      baseUrl?: string;
+      baseUrl: string;
```
Jakub-Vacek, line 72: I am not sure about the `baseUrl` change. We are counting on `serviceId.baseUrl` with fallback to `defaultService.baseUrl` now?

src/client/provider.ts
TheEdward162: We don't use it. The draft spec lets maps specify which service to use, so this would be redundant.

src/client/profile-provider.ts
TheEdward162: Yeah, `baseUrl` must be available and if it isn't then `bind` throws. I think we don't want to support absolute uris in maps for transparency sake.

src/client/registry.ts
TheEdward162: I don't think so? Ideally I'd like to use service client or something later.

Plus, this wasn't being used (maps are always retrieved using bind, even when Provider is local).

Although maybe that should change?

Jakub-Vacek: This will probably change with LUVs PR, right? U would just removed commented function 🤷

Jakub-Vacek (approved):

PR #140, opened 2021-09-28
feat: Add super.json schemas from ast

lukas-valenta:
## Description

Move super.json schema to ast-js. Depends on superfaceai/ast-js#63

## Motivation and Context

typescript-is generated validation is nicer to work with than zod and allows us to have a single source of truth

## Types of changes

- [x] None of the above

## Checklist:


src/internal/superjson/superjson.ts
```diff
@@ -102,7 +102,8 @@ export class SuperJson {
 
   static parse(input: unknown): Result<SuperJsonDocument, SDKExecutionError> {
     try {
-      const superdocument = superJsonSchema.parse(input);
+      // const superdocument = superJsonSchema.parse(input);
```
Jakub-Vacek, line 105: remove commented?

Jakub-Vacek (approved):

PR #147, opened 2021-10-04
fix: Flush metrics on unhandled rejection/exception

lukas-valenta:
## Description

When the user does not handle exceptions or promise rejections correctly, the metrics never get flushed as the process terminates. This adds handler for that, but it comes with the caveat that it consumes the event and never goes to default behavior. I'm not sure this is a good solution at all.

## Motivation and Context

We want to alert the backend about failure even if the user doesn't handle exceptions/rejections correctly.

## Types of changes

## Checklist:

### DIFF
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -154,7 +154,16 @@ export abstract class SuperfaceClientBase extends Events {
   }
 
   private hookMetrics(): void {
+    const unhandledHandler = (event: string) => (reason: unknown) => {
+      this.metricReporter?.flush();
+      // If the user has no defined handler, just print out the error similarly to default behaviour
+      if (process.rawListeners(event).length === 1) {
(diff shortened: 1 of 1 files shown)


src/client/client.ts
```diff
@@ -155,6 +155,16 @@ export abstract class SuperfaceClientBase extends Events {
 
   private hookMetrics(): void {
     process.on('beforeExit', () => this.metricReporter?.flush());
+    process.on('uncaughtExceptionMonitor', () => {
+      if (
+        process.rawListeners('uncaughtException').length === 0 &&
+        process.rawListeners('unhandledRejection').length === 0
```
TheEdward162, line 161: I think checking for `unhandledRejection` handlers here doesn't make sense. The exception could be coming from sync code.

PR #149, opened 2021-10-05
fix: Forbid absolute URLs in map

lukas-valenta:
## Description
This PR fixes two issues:
- it requires baseUrl to be specified for HTTP calls in map, as absolute URLs are no longer supported
- path parameter interpolation does not work on base URLs, only the relative path in map

## Motivation and Context

For security reasons, we need to only use base URL from provider.json and not allow interpolating it.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/src/internal/interpreter/http/http.ts
+++ b/src/internal/interpreter/http/http.ts
@@ -62,26 +62,21 @@ export enum NetworkErrors {
 
 export const createUrl = (
   inputUrl: string,
-  parameters?: {
-    baseUrl?: string;
+  parameters: {
+    baseUrl: string;
     pathParameters?: NonPrimitive;
   }
 ): string => {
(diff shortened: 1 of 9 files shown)


src/internal/interpreter/http/http.ts
```diff
@@ -62,27 +62,22 @@ export enum NetworkErrors {
...
-      throw new UnexpectedError(
-        'Relative URL specified, but base URL not provided!'
-      );
-    } else {
-      url = parameters.baseUrl.replace(/\/+$/, '') + inputUrl;
-    }
-  } else {
-    url = inputUrl;
+  if (!isRelative) {
+    console.log(inputUrl);
```
Jakub-Vacek, line 75: console.log

PR #172, opened 2021-11-02
fix: resolution of parameters with env variable

Jakub-Vacek:
## Description

This PR fixes resolution of integration parameters which are loaded from env

## Types of changes

## Checklist:

### DIFF
--- a/src/client/profile-provider.ts
+++ b/src/client/profile-provider.ts
@@ -3,7 +3,6 @@ import {
   assertProfileDocumentNode,
   FILE_URI_PROTOCOL,
   HttpScheme,
-  IntegrationParameter,
   isApiKeySecurityValues,
   isBasicAuthSecurityValues,
   isBearerTokenSecurityValues,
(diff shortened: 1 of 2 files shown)


martinalbert (approved):

martinalbert (approved):

src/client/profile-provider.ts
```diff
@@ -368,26 +368,34 @@ export class ProfileProvider {
...
-          // Use value as is
-        } else {
-          result[key] = value;
+      const providerJsonParameter = providerJsonParameters.find(
+        parameter => parameter.name === key
+      );
+      if (providerJsonParameter) {
+        //If value name and prepared value equals we are dealing with unset env
+        if (
+          prepareProviderParameters(providerJson.name, [providerJsonParameter])[
```
lukas-valenta, line 379: can you do this before the loop and just access it here so you don't call it (in worst case) every loop and also it will look better

lukas-valenta (approved):

PR #173, opened 2021-11-06
fix(ProfileParameterValidator): Resolve NamedModels first

lukas-valenta:
## Description

Changes ProfileParameterValidator behavior to resolve NamedModels before field references.

## Motivation and Context

When the field references were resolved before named models, the model definitions were not found, thus crashing the validator.

## Types of changes

## Checklist:

### DIFF
--- a/src/internal/interpreter/profile-parameter-validator.ts
+++ b/src/internal/interpreter/profile-parameter-validator.ts
@@ -483,12 +483,12 @@ export class ProfileParameterValidator implements ProfileVisitor {
     if (!this.namedDefinitionsInitialized) {
       node.definitions
         .filter(
-          (definition): definition is NamedFieldDefinitionNode =>
-            definition.kind === 'NamedFieldDefinition'
+          (definition): definition is NamedModelDefinitionNode =>
+            definition.kind === 'NamedModelDefinition'
         )
         .forEach(
           definition =>
-            (this.namedFieldDefinitions[definition.fieldName] = this.visit(
+            (this.namedModelDefinitions[definition.modelName] = this.visit(
               definition,
               kind,
               usecase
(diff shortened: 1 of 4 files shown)


src/internal/interpreter/profile-parameter-validator.ts
```diff
@@ -483,12 +483,12 @@ export class ProfileParameterValidator implements ProfileVisitor {
     if (!this.namedDefinitionsInitialized) {
       node.definitions
         .filter(
-          (definition): definition is NamedFieldDefinitionNode =>
-            definition.kind === 'NamedFieldDefinition'
+          (definition): definition is NamedModelDefinitionNode =>
+            definition.kind === 'NamedModelDefinition'
         )
```
martinalbert, line 488: Why not use profile ast utils? `isNamedModelDefinitionNode` and `isNamedFieldDefinitionNode` should work.

martinalbert, line 488: Why not use profile ast utils? `isNamedModelDefinitionNode` and `isNamedFieldDefinitionNode` should work.

martinalbert (commented):
I had similar problem resolving fields and models in profile-io-analyzer so I tested whether there won't be problem when I define field first and use it in model and it passes, but based on `visitFieldDefinitionNode()`, it checks whether [field is initialized in](https://github.com/superfaceai/one-sdk-js/blob/64c8b09eedc3c8faa28775c28bec70f46d2152d2/src/internal/interpreter/profile-parameter-validator.ts#L252-L256) `this.namedFieldDefinitions`, which will be always undefined when models are initialized first. 🤷🏻‍♂️

src/internal/interpreter/profile-parameter-validator.ts
lukas-valenta: i just moved the code, but good point :+1:

lukas-valenta:
i see, so it actually passes only because it's basically untyped :thinking: i'll think about it more and add some more tests

lukas-valenta:
@martinalbert I tried it out and it actually resolves correctly, because the values are validated lazily. Please try and find a counterexample if you think there's still a problem. In ProfileIOAnalyzer it's not resolved lazily, so there's a problem.

martinalbert (approved):
Yeah, I tried to debug it and you're right. 👍🏻

PR #174, opened 2021-11-11
feat: failover on bind error

Jakub-Vacek:
## Description

This PR adds ability to switch providers when bind is unsuccessful (related to https://www.notion.so/superface/Mails-are-not-send-from-Brain-7462ed7c349c45e8940749bbd0cd5c48#9bbf5ebde54e4dbaa5c54caa36fb6279). If the overall approach is ok I will add/fix tests and handle all resolution cases

## Types of changes

## Checklist:


TheEdward162 (commented):
I think this is fine

Jakub-Vacek:
I switched to bypassing the policy logic - in case of circuit breaker it would not switch. 
In case of error during bind we should switch to second provider if we can - providerFailover is true and we have second provider.

TheEdward162:
@Jakub-Vacek I think it would be better if the router could react to the error kind instead. So we'd add a new error variant to https://github.com/superfaceai/one-sdk-js/blob/d737f542aeb05fcda733e2d7dbdd95a4ee8b33f8/src/client/failure/policy.ts#L57 and then the policy would react to it by temporarily setting the number of retries to zero (e.g. switching right away). What do you think?

Jakub-Vacek:
Well, my solution is hacky, yours feels less hacky :D I will try it over the weekend.

Jakub-Vacek:
@TheEdward162 weekend is over (😅) and I tried to rewrite it according to your comment.

src/client/failure/policies.ts
```diff
@@ -264,6 +264,9 @@ export class RetryPolicy extends FailurePolicy {
   }
 
   override afterFailure(info: ExecutionFailure): FailureResolution {
+    if (info.kind === 'bind') {
+      this.streak = 0;
```
TheEdward162, line 268: Shouldn't this set the streak to maximum to force an abort?

Jakub-Vacek: Yep, i was handling this also in CircuitBreaker so this code never got used. Now it is handled in inner policy.

TheEdward162 (approved):

PR #175, opened 2021-11-12
Fix Buffer and Date cloning

janhalama:
## Description

This PR fixes `Buffer` and `Date` cloning

## Motivation and Context

ONESDK clones fetch params. If there is `Buffer` instance in body fetch parameter it is cloned as `Object` without prototype. Fetch then serialises the object as `[object Object]`.

This PR uses recursion to traverse over object properties and depending on the property type creates deep copy of it.

## Types of changes

## Checklist:

### DIFF
--- a/src/lib/object.test.ts
+++ b/src/lib/object.test.ts
@@ -66,6 +66,12 @@ describe('clone', () => {
     expect(cloned).toStrictEqual(null);
   });
 
+  it('should clone empty object', () => {
+    const object = {};
+    const cloned = clone(object);
+    expect(cloned).toStrictEqual({});
+  });
+
   describe('when cloning buffer', () => {
     let object: { buffer: Buffer };
 
--- a/src/lib/object.test.ts
+++ b/src/lib/object.test.ts
@@ -53,4 +53,79 @@ describe('clone', () => {
     const cloned = clone(object);
     expect(cloned).toStrictEqual(cloned);
   });
+
+  it('should clone undefined', () => {
+    const object = undefined;
+    const cloned = clone(object);
+    expect(cloned).toStrictEqual(undefined);
+  });
+
+  it('should clone null', () => {
+    const object = null;
+    const cloned = clone(object);
+    expect(cloned).toStrictEqual(null);
+  });
+
+  describe('when cloning buffer', () => {
+    let object: { buffer: Buffer };
+
+    beforeEach(() => {
+      object = {
+        buffer: Buffer.from('data'),
+      };
+    });
+
+    it('should clone buffer', () => {
+      const cloned = clone(object);
+      expect(Buffer.isBuffer(cloned.buffer)).toBe(true);
+      expect(cloned.buffer.toString()).toEqual(object.buffer.toString());
+    });
+
+    it('should create new instance of buffer', () => {
+      const cloned = clone(object);
+      expect(cloned.buffer !== object.buffer).toBe(true);
+    });
+  });
+
+  describe('when cloning array', () => {
+    let object: { array: Array<any> };
+
+    beforeEach(() => {
+      object = {
+        array: [1, { a: 1, b: 2, c: 'string' }],
+      };
+    });
+
+    it('should clone array', () => {
+      const cloned = clone(object);
+      expect(Array.isArray(cloned.array)).toBe(true);
+      expect(cloned).toStrictEqual(object);
+    });
+
+    it('should create new instance of array', () => {
+      const cloned = clone(object);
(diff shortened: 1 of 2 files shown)


src/lib/object.ts
```diff
@@ -4,7 +4,43 @@ import { UnexpectedError } from '../internal/errors';
...
+    return arrayCopy.map((item: unknown) =>
+      clone<unknown>(item)
+    ) as unknown as T;
+  }
+
+  if (Buffer.isBuffer(value)) {
+    return Buffer.from(value) as unknown as T;
+  }
+
+  if (typeof value === 'object' && value !== {}) {
```
jnv, line 30: What is the intent with this condition: `value !== {}`? It never evaluates to false since you are comparing references. If you want to check for an empty object, you will need something like `Object.keys(value).length > 0`

jnv: However I would dropping this condition altogether, I think it's an unnecessary optimization, the code will still work with empty objects (though make sure to test it).

src/lib/object.ts
```diff
@@ -4,7 +4,43 @@ import { UnexpectedError } from '../internal/errors';
...
+
+  if (value instanceof Array) {
+    const arrayCopy = [] as unknown[];
+    value.forEach(item => {
+      arrayCopy.push(item);
+    });
+
+    return arrayCopy.map((item: unknown) =>
+      clone<unknown>(item)
+    ) as unknown as T;
```
jnv, line 23: Can you avoid iterating the array twice? e.g.
```suggestion
   for (const item of value) {
     arrayCopy.push(clone<unknown>(item))
   }

   return arrayCopy as unknown as T;
```

(it's a pretty minimal optimization but for/of typically beats forEach performance without sacrificing readability)

lukas-valenta: i think creating the copy array is unnecessary, something like `return value.map(clone)` could be enough, but typescript will probably shout about it

src/lib/object.ts
```diff
@@ -4,7 +4,43 @@ import { UnexpectedError } from '../internal/errors';
...
-  return JSON.parse(JSON.stringify(value)) as T;
+  if (value === null) {
+    return value;
+  }
+
+  if (value instanceof Date) {
+    return new Date(value.getTime()) as unknown as T;
+  }
+
+  if (value instanceof Array) {
```
lukas-valenta, line 15:
```suggestion
  if (Array.isArray(value)) {
```

src/lib/object.ts
janhalama: @jnv you are right the condition for empty object was wrong. I removed it and added test.

janhalama:
@jnv @lukas-valenta thanks for review, I fixed your comments, 🙏 review once more

src/lib/object.ts
```diff
@@ -4,7 +4,41 @@ import { UnexpectedError } from '../internal/errors';
...
+      ...(value as unknown as { [key: string]: unknown }),
+    } as {
+      [key: string]: unknown;
+    };
+    Object.keys(objectCopy).forEach(key => {
+      objectCopy[key] = clone<unknown>(objectCopy[key]);
+    });
+
+    return objectCopy as unknown as T;
+  }
```
lukas-valenta, line 39:
```suggestion
  if (typeof value === 'object') {
    const objectCopy = Object.entries(value).map(([key, value]) => [
      key,
      clone(value) as unknown,
    ]);

    return Object.fromEntries(objectCopy) as unknown as T;
  }
```

i think this looks a bit more neat

lukas-valenta (approved):
looks good, just please run format before merging

src/lib/object.test.ts
```diff
@@ -53,4 +53,85 @@ describe('clone', () => {
...
+
+    it('should clone array', () => {
+      const cloned = clone(object);
+      expect(Array.isArray(cloned.array)).toBe(true);
+      expect(cloned).toStrictEqual(object);
+    });
+
+    it('should create new instance of array', () => {
+      const cloned = clone(object);
+      expect(cloned.array !== object.array).toBe(true);
```
jnv, line 113: I recommend using Jest's built-in expectation for checking referential inequality, it's shorter and more readable:

```suggestion
      expect(cloned).not.toBe(object);
```

Alternatively, since we don't care about deep comparison here, per [docs](https://jestjs.io/docs/expect#tobevalue) you may want to use `Object.is` directly (which is preferred over `===`):

```suggestion
      expect(Object.is(object, cloned)).toBe(false)
```

Same applies to lines 92 and 134.

PR #179, opened 2021-11-17
feat: digest auth

Jakub-Vacek:
## Description

This PR adds digest authentication. Missing some tests and more importantly `user` and `password` inputs. Tested against real ClockPlus API.

## Types of changes

## Checklist:


src/internal/interpreter/http/digest.ts
```diff
@@ -0,0 +1,171 @@
...
+    if (match) return trim ? match[1].replace(/[\s"]/g, '') : match[1];
+
+    return undefined;
+  }
+
+  private static computeHash(
+    user: string,
+    realm: string,
+    password: string
+  ): string {
```
TheEdward162, line 166: If I understand it correctly this is supposed to be the default route for when no algorithm is specified? The name is vague and the algorithm is harcoded, so I'm not sure.

src/internal/interpreter/http/digest.ts
```diff
@@ -0,0 +1,262 @@
...
+    private readonly cnonceSize: number = 32,
+    //407 can be also used when communicating thru proxy https://datatracker.ietf.org/doc/html/rfc2617#section-1.2
+    private readonly statusCode: number = 401,
+    //"Proxy-Authenticate" can be also used when communicating thru proxy https://datatracker.ietf.org/doc/html/rfc2617#section-1.2
+    private readonly header: string = 'www-authenticate',
+    private nc: number = 0
+  ) {
+    debug('Initialized DigestHelper');
+    debugSensitive(
+      `Initialized with: userame="${this.user}", password="${this.password}", cnonce size=${this.cnonceSize}, status code=${this.statusCode}, challenge header="${this.header}"`
```
jnv, line 54: Typo: userame

src/internal/interpreter/http/http.ts
```diff
@@ -170,101 +234,91 @@ export class HttpClient {
       }
 
       if (configuration.type === SecurityType.APIKEY) {
-        applyApiKeyAuth(contextForSecurity, configuration);
+        const handler = new ApiKeyHandler()
+        handler.prepare(contextForSecurity, configuration)
```
lukas-valenta, line 238: if you don't need to do stuff with the constructed unprepared handler, it'll be probably better to use static prepare for construction, i.e. `const handler = ApiKeyHandler.prepare(contextForSecurity, configuration)`

src/client/profile-provider.ts
```diff
@@ -67,11 +67,12 @@ function profileAstId(ast: ProfileDocumentNode): string {
 const boundProfileProviderDebug = createDebug(
   'superface:bound-profile-provider'
 );
+export type AuthCache = { cache?: { digest?: string } };
```
lukas-valenta, line 70: shouldn't there be some sort of cache key?

Jakub-Vacek: This exists on boundProfileProvider so it is already scoped only on one provider. Not sure if something more is needed:
 - if user reuses boundProfileProvider with different auth. method this won't be used

src/internal/interpreter/http/http.ts
Jakub-Vacek: You mean the `ApiKeyHandler.prepare(contextForSecurity, configuration)` should change the `contextForSecurity` and return new `ApiKeyHandler` instance? It feels like two different things. But maybe the "change the `contextForSecurity`" part is hard to spot at first 🤔

lukas-valenta: oh i didn't notice it changes the context. that's not not very good tbh :thinking:

lukas-valenta: but it's okay now, with the configuration in constructor :shrug:

src/internal/interpreter/http/security/digest/digest.ts
```diff
@@ -0,0 +1,296 @@
...
+  //Internal
+  private readonly nonceRaw: string = 'abcdef0123456789';
+  private readonly cnonceSize = 32;
+  private nc = 0;
+
+  constructor(
+    readonly configuration: DigestSecurityScheme & DigestSecurityValues
+  ) {
+    debug('Initialized DigestHelper');
+    this.statusCode = configuration.statusCode || 401;
```
TheEdward162, line 59: Can we has `??` here?

src/internal/interpreter/http/security/digest/digest.ts
```diff
@@ -0,0 +1,296 @@
...
+      qop: extractQop(header),
+      nonce,
+      cnonce: this.makeNonce(),
+    };
+  }
+
+  private makeNonce(): string {
+    let uid = '';
+    for (let i = 0; i < this.cnonceSize; ++i) {
+      uid += this.nonceRaw[Math.floor(Math.random() * this.nonceRaw.length)];
```
TheEdward162, line 203: I'm not sure about the `Math.random` security implications here.

@lukas-valenta What do you think?

TheEdward162: On the other hand, it is only used to compute cnonce.. 🤷

src/internal/interpreter/http/security/digest/digest.ts
```diff
@@ -0,0 +1,296 @@
...
+      ]);
+    }
+  }
+
+  // when not specified
+  return undefined;
+}
+
+function extract(raw: string, field: string, trim = true): string | undefined {
+  //TODO: field should be case-insensitive
```
TheEdward162, line 273: This todo no longer applies, right?

src/lib/fetch.ts
```diff
@@ -65,9 +65,10 @@ export class CrossFetch implements FetchInstance, Interceptable {
     let body: unknown;
 
     if (
-      (headers['content-type'] &&
-        headers['content-type'].includes(JSON_CONTENT)) ||
-      parameters.headers?.['accept']?.includes(JSON_CONTENT)
+      headers['content-type'] &&
+      headers['content-type'].includes(JSON_CONTENT) //||
+      //This should be used when we don't have a response content-type
```
TheEdward162, line 70: Please make this a TODO for when security handlers get updated to make requests themselves, as we talked.

src/internal/interpreter/http/security/digest/digest.ts
lukas-valenta: i missed that. cnonce or no, let's use crypto

Jakub-Vacek:
@lukas-valenta @TheEdward162 Technically, this is good to go - it does not depend on any rc/beta releases, it works and it is covered with tested. Can we merge it or do we want wait on events auth. architecture? 🤷🏻‍♂️

jnv:
Please land it before it drifts away from main. 😅

lukas-valenta:
@jnv too late for that :disappointed:

lukas-valenta (approved):

PR #191, opened 2021-12-22
feat: Allow passing parameters to perform

lukas-valenta:
## Description
Allows overriding integration parameters in perform call

## Motivation and Context
This allows us to configure specific calls at runtime, like per-user configuration

## Types of changes

## Checklist:

### DIFF
--- a/src/client/profile-provider.ts
+++ b/src/client/profile-provider.ts
@@ -69,7 +69,7 @@ const boundProfileProviderDebug = createDebug(
 );
 
 export class BoundProfileProvider {
-  //TODO: Interceptable and set metadata
+  // TODO: Interceptable and set metadata
   private profileValidator: ProfileParameterValidator;
   private fetchInstance: FetchInstance & Interceptable;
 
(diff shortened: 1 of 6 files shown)


src/client/profile-provider.ts
```diff
@@ -181,6 +185,24 @@ export class BoundProfileProvider {
 
     return ok(result.value);
   }
+
+  private mergeParameters(
+    parameters?: Record<string, string>,
+    providerParameters?: Record<string, string>
+  ): Record<string, string> {
+    if (parameters === undefined) {
+      return providerParameters ?? {};
```
martinalbert, line 194: Why not possibly return undefined? Map interpreter takes parameters as optional parameter

martinalbert (approved):

src/client/profile-provider.ts
lukas-valenta: you're right. to be honest i didn't write this code, copilot did :grin:

martinalbert: oh nice 😄 could it write it correctly if you add undefined as return type too?

lukas-valenta: hard to say, it wrote the return type itself :D

PR #194, opened 2022-01-10
Feature/service selection

TheEdward162:
## Description
Service selection should now work.

## Motivation and Context
We added service selection to ast and parser, now we need to support it in sdk.

## Types of changes

## Checklist:

### DIFF
--- a/src/internal/interpreter/profile-parameter-validator.ts
+++ b/src/internal/interpreter/profile-parameter-validator.ts
@@ -141,6 +141,8 @@ export class ProfileParameterValidator implements ProfileVisitor {
         return this.visitComlinkObjectLiteralNode(node, kind, usecase);
       case 'ComlinkPrimitiveLiteral':
         return this.visitComlinkPrimitiveLiteralNode(node, kind, usecase);
+      case 'ComlinkAssignment':
+        throw new Error("unreachable");
       case 'EnumDefinition':
         return this.visitEnumDefinitionNode(node, kind, usecase);
       case 'EnumValue':
--- a/src/internal/interpreter/profile-parameter-validator.ts
+++ b/src/internal/interpreter/profile-parameter-validator.ts
@@ -1,4 +1,5 @@
 import {
+  ComlinkAssignmentNode,
   ComlinkListLiteralNode,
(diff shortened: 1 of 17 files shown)


src/internal/interpreter/profile-parameter-validator.ts
```diff
@@ -141,6 +141,8 @@ export class ProfileParameterValidator implements ProfileVisitor {
         return this.visitComlinkObjectLiteralNode(node, kind, usecase);
       case 'ComlinkPrimitiveLiteral':
         return this.visitComlinkPrimitiveLiteralNode(node, kind, usecase);
+      case 'ComlinkAssignment':
+        throw new Error('unreachable');
```
lukas-valenta, line 145: do this properly as it is in other node types

src/lib/services.ts
```diff
@@ -0,0 +1,31 @@
+import { ProviderService } from '@superfaceai/ast';
+
+export class ServicesSelector {
```
lukas-valenta, line 3: 1. can this just be ServiceSelector? ServicesSelector triggers me in unexpected ways
2. can you make an IServiceSelector for this with just the `getUrl` method and let this implement it (and interpreter consume just the interface)?

src/lib/services.ts
```diff
@@ -0,0 +1,31 @@
...
+   * Gets the url of `service`. If `service` is undefined returns url of the default service, or undefined if default service is also undefined.
+   */
+  getUrl(service?: string): string | undefined {
+    const srvc = service ?? this.defaultService;
+    if (srvc === undefined) {
+      return undefined;
+    }
+
+    return this.serviceUrls[srvc];
+  }
```
lukas-valenta, line 30:
```suggestion
  getUrl(serviceId?: string): string | undefined {
    const service = serviceId ?? this.defaultService;
    if (service === undefined) {
      return undefined;
    }

    return this.serviceUrls[service];
  }
```

lukas-valenta (changes requested):
looks good, some minor nitpicks. good to merge after dependencies release

lukas-valenta (approved):

PR #217, opened 2022-03-08
Feat/bound provider timeout

Jakub-Vacek:
## Types of changes

## Checklist:

### DIFF
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -36,7 +36,7 @@ export abstract class SuperfaceClientBase extends Events {
   public readonly superJson: SuperJson;
   private readonly metricReporter: MetricReporter | undefined;
   private boundCache: {
-    [key: string]: BoundProfileProvider;
+    [key: string]: { profileProvider: BoundProfileProvider; expiresAt: number };
   } = {};
 
   public hookContext: HooksContext = {};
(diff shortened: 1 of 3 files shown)


lukas-valenta (approved):

src/client/client.ts
```diff
@@ -73,7 +73,8 @@ export abstract class SuperfaceClientBase extends Events {
     const cacheKey = profileConfig.cacheKey + providerConfig.cacheKey;
 
     const bound = this.boundCache[cacheKey];
-    if (bound === undefined) {
+    const now = Math.floor(Date.now() / 1000);
+    if (bound === undefined || bound.expiresAt < now) {
```
lukas-valenta, line 77: Now that I think about it, let's do this as @jnv suggested:
> As a sort of compromise you could do a background rebinding on demand (similar to Next.js Incremental Static Regeneration if you are familiar)
> - Perform comes in with provider already bound
> - Do perform with the current (bound) version of map
> - Check if TTL expired, if yes, schedule a rebinding (on nextTick or something similar)
> - While rebinding happens, use the old version of map
> - Once rebind finishes, swap map with the new version

lukas-valenta (changes requested):

src/client/client.ts
Jakub-Vacek: Ok, i will start working on it next week :)

lukas-valenta (approved):

PR #218, opened 2022-03-09
feat: Cache provider.json

lukas-valenta:
## Description

Adds naive `provider.json` caching

## Motivation and Context

We are caching maps to be able to fallback if bind call fails, but provider isn’t cached, so the map has no use. Now if the fetch fails, loading from cache is attempted.

## Types of changes

## Checklist:


src/client/profile-provider.ts
```diff
@@ -417,6 +421,63 @@ export class ProfileProvider {
...
+      const providerCachePath = joinPath(cachePath, `${providerName}.json`);
+
+      // If we don't have provider info, we first try to fetch it from the registry
+      try {
+        this.providerJson = await fetchProviderInfo(providerName);
+        try {
+          await fsp.mkdir(cachePath, { recursive: true });
+          await fsp.writeFile(
+            providerCachePath,
+            JSON.stringify(this.providerJson)
```
Jakub-Vacek, line 437: Nitpick: nice formatting? (JSON.stringify(this.providerJson, undefined,2))

src/client/profile-provider.test.ts
```diff
@@ -47,6 +47,8 @@ jest.mock('fs', () => ({
   ...jest.requireActual<Record<string, unknown>>('fs'),
   promises: {
     readFile: jest.fn(),
+    mkdir: jest.fn(),
```
Jakub-Vacek, line 50: No more tests?

src/client/profile-provider.ts
lukas-valenta: it's not really meant to be read :thinking:

src/client/profile-provider.test.ts
lukas-valenta: I'd spend an hour mocking everything and then just delete it in refactor when I'm rewriting it, so I don't think the value is there :shrug:

src/client/profile-provider.ts
Jakub-Vacek: Yeah, I know. I was just an idea. It would be nice for people who decide to dive deeper to superface.

src/client/profile-provider.test.ts
Jakub-Vacek: Hmm, hour of mocking sucks but it is not cool to not have tests 😩
I am ok with writing them (Friday or Monday) but we need to break this file (during your refactor) into more testable pieces.

lukas-valenta: I added tests. Hope you're happy.

Jakub-Vacek (approved):

PR #221, opened 2022-03-16
Feat/pass super json as parameter

Jakub-Vacek:
This PR adds option to pass SuperJson or SuperJsonDocument to SuperfaceClient constructor. This parameter is used instead of local super.json

## Motivation and Context

https://www.notion.so/superface/SuperJson-as-SDK-parameter-aa2462ab4bb24239b4283cbb34fcec74

## Types of changes

## Checklist:


src/client/client.ts
```diff
@@ -41,15 +43,24 @@ export abstract class SuperfaceClientBase extends Events {
...
+        options.superJson instanceof SuperJson
+          ? options.superJson
+          : new SuperJson(options.superJson);
+    }
     const superCacheKey = Config.instance().superfacePath;
 
     if (SUPER_CACHE[superCacheKey] === undefined) {
-      SUPER_CACHE[superCacheKey] = SuperJson.loadSync(superCacheKey).unwrap();
+      SUPER_CACHE[superCacheKey] =
+        passedSuperJson ?? SuperJson.loadSync(superCacheKey).unwrap();
```
lukas-valenta, line 59: suggestion: this could be extracted to function to just do `SUPER_CACHE[superCacheKey] = getSuperJson(options?.superJson)` to keep the constructor neat.

lukas-valenta: also the caching should probably be done differently for passed parameter :thinking:

PR #222, opened 2022-03-18
Feat/pass security values

Jakub-Vacek:
This PR adds option to pass security values directly to perform call. Now it is more like POC of how it can be done, if we need to merge passed values with security values from super.json and how hard it will be to use this.

## Types of changes

## Checklist:

### DIFF
--- a/src/client/usecase.ts
+++ b/src/client/usecase.ts
@@ -59,6 +59,18 @@ class UseCaseBase implements Interceptable {
         `${this.profile.configuration.id}/${this.name}`
       ].router;
 
+    //In this instance we can set metadata for events
+    this.boundProfileProvider =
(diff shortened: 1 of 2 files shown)


src/client/usecase.ts
```diff
@@ -76,6 +77,13 @@ class UseCaseBase implements Interceptable {
     this.metadata.provider = providerConfig.name;
     hookRouter.setCurrentProvider(providerConfig.name);
 
+    //If we have security values pass them directly to perform
+    if (options?.security) {
+      providerConfig = new ProviderConfiguration(
```
lukas-valenta, line 82: suggestion: once again, I would extract this to a function `const providerConfig = getProviderConfiguration(options)`

lukas-valenta (approved):

PR #229, opened 2022-05-07
fix: http::replaceParameters now trims whitespace

### LINKED ISSUE #228: [BUG] Path parameters with spaces around
Provide a general summary of the issue in the Title above

Unable to match parameters in path when passed like this: 
```
  pathParams = {
    city : "Prague",
  }
  http GET "/{ pathParams.city }" {
```
Spaces around are the problem.

## Expected Behavior
It should match the `city` parameter 

## Current Behavior
It does not match `city` parameter and end with error (with correct hint)

## Possible Solution
Trim the parameter name in SDK interpreter or do not allow spaces around path parameters (in parser)

## Steps to Reproduce
Just wrap name of the path parameter in spaces

TheEdward162:
## Description
Rewritten `replaceParameters` using the newer `String.matchAll` api while taking care of whitespace. I believe this makes the method more readable as well.

## Motivation and Context
Fixes #228 

## Types of changes

## Checklist:

### DIFF
--- a/src/internal/interpreter/http/http.ts
+++ b/src/internal/interpreter/http/http.ts
@@ -79,7 +79,7 @@ function replaceParameters(url: string, parameters: NonPrimitive) {
     const start = match.index;
     // Why can this be undefined?
     if (start === undefined) {
-      throw new Error('Unexpected regex match state - missing start index');
+      throw new UnexpectedError('Invalid regex match state - missing start index');
     }
 
     const end = start + match[0].length;
--- a/src/internal/interpreter/http/http.ts
+++ b/src/internal/interpreter/http/http.ts
@@ -18,6 +18,7 @@ import {
   NonPrimitive,
   Variables,
   variablesToStrings,
+  variableToString,
 } from '../variables';
 import {
   authenticateFilter,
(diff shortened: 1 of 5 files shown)


Jakub-Vacek (approved):
👍

src/internal/interpreter/http/http.ts
```diff
@@ -67,38 +68,43 @@ export enum NetworkErrors {
...
   const regex = RegExp('{([^}]*)}', 'g');
-  let replacement: RegExpExecArray | null;
-  while ((replacement = regex.exec(url)) !== null) {
-    replacements.push(replacement[1]);
-  }
+  for (const match of url.matchAll(regex)) {
+    const start = match.index;
+    // Why can this be undefined?
+    if (start === undefined) {
+      throw new Error('Unexpected regex match state - missing start index');
```
lukas-valenta, line 82: suggestion: use UnexpectedError

src/internal/interpreter/variables.ts
```diff
@@ -109,6 +109,19 @@ export const getValue = (
   return result;
 };
 
+/**
+ * Stringifies a primitive variable.
+*/
+export const variableToString = (
```
lukas-valenta, line 115: suggestion: It may not be clear what exactly this does because the comment says primitive, but it can take anything. Could we either make it only take `string | boolean | number` or explain it in the comment?

src/internal/interpreter/variables.ts
```diff
@@ -120,7 +133,7 @@ export const variablesToStrings = (
   if (variables) {
     for (const [key, value] of Object.entries(variables)) {
       if (value !== undefined) {
-        result[key] = typeof value === 'string' ? value : JSON.stringify(value);
+        result[key] = variableToString(value);
```
lukas-valenta, line 136: praise: It is much nicer this way :+1:

src/internal/interpreter/map-interpreter.test.ts
```diff
@@ -228,6 +228,40 @@ describe('MapInterpreter', () => {
     expect(result.isOk() && result.value).toEqual(144);
   });
 
+  it('should correctly trim and parse path parameters in http call', async () => {
```
lukas-valenta, line 231: todo: in the future it would be better to export the function and test it directly

lukas-valenta (approved):

src/internal/interpreter/variables.ts
TheEdward162: You are right, the comment is wrong. I'm not sure why I wrote "primitive".

TheEdward162: Should we try to handle objects/arrays specially? It doesn't seem to be a problem right now but maybe we are just missing a bug here.

src/internal/interpreter/map-interpreter.test.ts
TheEdward162: You are right, I added the higher-level test because that's how I confirmed the bug reproduction and didn't think to test it at a lower level. I've added a test into `http.test.ts` as well.

src/internal/interpreter/variables.ts
lukas-valenta: I'm not sure. It worked like this so far, so let's leave it until we have a case for changing it :shrug:

Issue #230, opened 2022-05-12
[BUG] Not possible to send form-data with duplicate key names


jnv:
Based on https://github.com/superfaceai/station/issues/242

Mailgun in particular expects email attachments to be submitted under `attachment` fields in `multipart/form-data`. However, we don't allow duplicated keys to be sent in body, because it is treated as a JS object.

Possible workaround would be to detect array in value when constructing the FormData object, and append each value individually.

https://github.com/superfaceai/one-sdk-js/blob/1f03560419ab141382d02b5cc51e17aa14365ca4/src/lib/fetch.ts#L192-L194

This makes sense because there isn't any official way to serialize array values in FormData.

Therefore the following Comlink:

```
body {
  attachment = [bufferA, bufferB]
}
```

Results in:

```
attachment=bufferA
attachment=bufferB
```

PR #231, opened 2022-05-12
fix(http): Don't set content-type w/ form-data

jnv:
## Description

Related to #230 (but not root cause).

When I want to send body with `multipart/form-data` encoding to Mailgun, I am getting back server error "Missing content boundary". Upon closer inspection, the [boundary directive](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type#directives) is missing in the `Content-Type` header. It is normally automatically read from `FormData`, but since `content-type` is set explicitly, it's not overwritten.

Thanks to @lukas-valenta for helping me to solve this issue.

## Types of changes

## Checklist:

### DIFF
--- a/src/internal/interpreter/http/http.test.ts
+++ b/src/internal/interpreter/http/http.test.ts
@@ -150,4 +150,32 @@ describe('HttpClient', () => {
       });
     }
   });
+
+  describe('multipart/form-data', () => {
+    it('encodes multipart boundary correctly', async () => {
+      await mockServer.post('/data').thenCallback(async req => {
+        return {
+          status: 200,
+          headers: { 'content-type': 'application/json' },
+          body: JSON.stringify({
+            contentType: req.headers['content-type'],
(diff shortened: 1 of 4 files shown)


src/internal/interpreter/http/http.test.ts
```diff
@@ -138,4 +138,33 @@ describe('HttpClient', () => {
...
+      const response = await http.request('/data', {
+        method: 'post',
+        contentType: 'multipart/form-data',
+        body: {
+          foo: 1,
+          bar: 'baz',
+        },
+        baseUrl,
+      });
+      console.log(response);
```
lukas-valenta, line 164:
```suggestion
```

lukas-valenta (approved):

jnv:
@lukas-valenta I have recreated the PR post-release, as discussed.

lukas-valenta (approved):

PR #236, opened 2022-05-30
Refactor/client separation

lukas-valenta:
## Description
Refactoring to improve inversion of control and other stuff

## Types of changes

## Checklist:


src/client/index.ts
```diff
@@ -1,5 +1,7 @@
-export { SuperfaceClient, createTypedClient, typeHelper } from './client';
-export { Profile, ProfileConfiguration, TypedProfile } from './profile';
+// export { SuperfaceClient, createTypedClient, typeHelper } from './client';
+export { SuperfaceClient } from './client';
+// export { Profile, ProfileConfiguration, TypedProfile } from './profile';
```
Jakub-Vacek, line 3: Do you need to keep the commented exports?

Jakub-Vacek:
Overall I like it 👍

I would probably ( (k)nitpicking, not saying we should do it now ):
- separate thing like `Profile` and `TypedProfile` to separate files to shorten and simplify tests. 
- make base classes abstract if possible (maybe you use them in tests for mocking)
- think about using SDK without filesystem - not needed now but we will want it in the future

I will probably come up with few more ides but it won't be anything serious. Really nice work 🥳🙏

src/client/cache.ts
```diff
@@ -0,0 +1,37 @@
...
+    const cached = this.cache[cacheKey];
+    if (cached !== undefined) {
+      return cached;
+    }
+
+    const initialized = initializer();
+    if (initialized instanceof Promise) {
+      return new Promise((resolve, reject) => {
+        initialized
+          .then(value => {
```
TheEdward162, line 19: question: Does this have to create a new Promise? Can't it just be
```
initialized.then(value => {
  this.cache[cacheKey] = value;
  return value;
})
```
?

src/client/client.internal.ts
```diff
@@ -0,0 +1,75 @@
...
+      this.events,
+      this.superJson,
+      this.config,
+      this.fileSystem,
+      this.boundProfileProviderCache,
+      this.logger
+    );
+  }
+
+  public async getProfileConfiguration(
```
TheEdward162, line 42: nitpick: `public` is explicitly used here but only implicitly used above in `getProfile`.

src/client/client.ts
```diff
@@ -1,235 +1,108 @@
...
-
-    return new Provider(
-      this,
-      new ProviderConfiguration(providerName, providerSettings.security)
-    );
+    return getProvider(this.superJson, providerName);
   }
 
   /** Returns a provider configuration for when no provider is passed to untyped `.perform`. */
   async getProviderForProfile(profileId: string): Promise<Provider> {
```
TheEdward162, line 90: nitpick: I can probably `git blame` this all the way back to myself, but maybe we should rename this to be more descriptive. Like `getDefaultProviderForProfile` or something like that 🤷‍♀️

lukas-valenta: I did not go into breaking changes yet

src/client/cache.ts
lukas-valenta: good question

TheEdward162 (commented):
First part of my review, more files pending

Jakub-Vacek:
Why do we resolve/read profile file during `bind`? I feel it should be done sooner (during `getProfile`).

src/client/usecase.ts
```diff
@@ -76,12 +101,35 @@ class UseCaseBase implements Interceptable {
...
+          this.config,
+          this.events,
+          this.fileSystem,
+          this.logger
+        )
       );
+    const now = Math.floor(Date.now() / 1000);
+    if (expiresAt < now) {
+      this.boundProfileProviderCache.invalidate(cacheKey);
+      void this.rebind(cacheKey, providerConfig);
```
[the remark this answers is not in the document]
lukas-valenta: The idea was that after expiration we use the cached version and fetch the new one in the meantime to be ready for the next perform. There's probably a DR in Slack or Notion

src/lib/reporter.ts
```diff
@@ -134,6 +135,50 @@ export type EventInput =
   | PerformMetricsInput
   | ProviderChangeInput;
 
+export function hookMetrics(
+  events: Events,
+  metricReporter: MetricReporter
+): void {
+  process.on('beforeExit', () => metricReporter?.flush());
```
TheEdward162, line 142: question: Why the one-eyed elvis `?.`?

src/lib/logger/logger.ts
```diff
@@ -0,0 +1,9 @@
+export type LogFunction = {
+  (format: string, ...args: unknown[]): void;
+  enabled: boolean;
+};
+
+export interface ILogger {
+  log(name: string): LogFunction;
+  log(name: string, format: string, ...args: unknown[]): void;
+}
```
TheEdward162, line 9: praise: Meesa likey abstract logger

src/lib/environment/environment.node.ts
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  public getBoolean(key: string): boolean | undefined {
+    const value = process.env[key]?.trim();
+
+    if (value === undefined) {
+      return undefined;
+    }
+
+    return value.toLowerCase() === 'true' || value === '1';
```
TheEdward162, line 25: troll: bUt WhAt If It IS An UPPeRcAsE oNe?

src/config.ts
```diff
@@ -1,7 +1,23 @@
...
-
-const configDebug = createDebug('superface:config');
+import { IEnvironment } from './lib/environment';
+import { IFileSystem } from './lib/io';
+import { NodeFileSystem } from './lib/io/filesystem.node';
+import { ILogger, LogFunction } from './lib/logger/logger';
+
+const DEBUG_NAMESPACE = 'config';
+
+type JoinPath = { path: { join: IFileSystem['path']['join'] } };
```
TheEdward162, line 8: question: Why not just request `IFileSystem` directly?

lukas-valenta: Because this is used outside of SDK and having other users implement the entire FileSystem seems like overkill.. but the filesystem dependency will be removed soon :tm: anyway hopefully

src/config.ts
```diff
@@ -5,7 +5,9 @@ import { ILogger, LogFunction } from './lib/logger/logger';
 
 const DEBUG_NAMESPACE = 'config';
 
-type JoinPath = { path: { join: IFileSystem['path']['join'] } };
+type FSPath = {
+  path: { join: IFileSystem['path']['join']; cwd: () => string };
+};
```
TheEdward162, line 10: suggestion: `type FSPath = { path: Pick<IFileSystem['path'], 'join' | 'cwd'> };`

src/internal/superjson/schema.ts
```diff
@@ -3,18 +3,16 @@ import {
...
     return path;
   }
-  const normalizedPath = normalize(path);
 
-  return normalizedPath.startsWith('../')
-    ? `${FILE_URI_PROTOCOL}${normalizedPath}`
-    : `${FILE_URI_PROTOCOL}./${normalizedPath}`;
+  return path.startsWith('../')
+    ? `${FILE_URI_PROTOCOL}${path}`
+    : `${FILE_URI_PROTOCOL}./${path}`;
```
TheEdward162, line 17: question: since this doesn't normalize anymore (I assume because normalization now comes through the `IFilesystem`) shouldn't we normalize at the call sites? Is it possible to correctly emulate the original behavior (i.e. ignore already-prefixed paths) externally? I don't remember why exactly this was needed in the first place

lukas-valenta: I did normalize at call sites

TheEdward162: How about these (and more)?

https://github.com/superfaceai/one-sdk-js/blob/cea3a9e80ec0165ce05fc9a27a47be50599051ec/src/internal/superjson/mutate.ts#L138

https://github.com/superfaceai/one-sdk-js/blob/cea3a9e80ec0165ce05fc9a27a47be50599051ec/src/internal/superjson/mutate.ts#L652

TheEdward162: 🍆

src/client/usecase.ts
```diff
@@ -67,12 +95,38 @@ class UseCaseBase implements Interceptable {
...
+      this.profileConfiguration.cacheKey + providerConfig.cacheKey,
+      providerConfig
+    );
+  }
+
+  private async rebind(
+    cacheKey: string,
+    providerConfig: ProviderConfiguration
+  ): Promise<IBoundProfileProvider> {
+    console.log(cacheKey, providerConfig);
```
TheEdward162, line 108: question: What's up with this bad boy?

src/client/profile-provider.ts
```diff
@@ -627,39 +562,43 @@ export class ProfileProvider {
...
+          contents = await fileSystem.readFile(fileNameWithExtension);
+          if (contents.isOk()) {
             break;
-          } catch (e) {
-            void e;
           }
         }
 
-        if (contents === undefined) {
+        if (contents === undefined || contents.isErr()) {
```
Jakub-Vacek, line 584: We were `void e` before now we throw `referencedFileNotFoundError`. Not sure if this change is intentional and if we really want to throw referencedFileNotFoundError 🤷🏻‍♂️

lukas-valenta: read it again carefully, the behaviour did not change

src/internal/superjson/mutate.ts
```diff
@@ -233,7 +256,14 @@ export function setProfile(
 
   // if payload is undefined we already deleted it (or it wasn't present)
   if (payload !== undefined) {
-    const mergeChanged = mergeProfile(document, profileName, payload);
+    const mergeChanged = mergeProfile(
```
Jakub-Vacek, line 259: nitpick: these could be passed as an object.

src/core/interfaces/filesystem.errors.ts
```diff
@@ -0,0 +1,46 @@
+export class FileExistsError extends Error {
```
Jakub-Vacek, line 1: I would expect this file in errors folder 🤷🏻‍♂️

Jakub-Vacek: Or like the error definitions should be there and `FileSystemError` maybe here.

tsconfig.json
```diff
@@ -1,14 +1,25 @@
...
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "forceConsistentCasingInFileNames": true,
-    "lib": [
-      "ESNext"
-    ],
+    "lib": ["ESNext"],
+    "paths": {
+      "~core": ["core"],
+      "~core/*": ["core/*"],
```
TheEdward162, line 13: kinder: 😮 Why didn't we know about this before

TheEdward162 (approved):
I also like the directory structure/module separation.

Jakub-Vacek (approved):

PR #252, opened 2022-07-04
feat: early remote profile resolution

Jakub-Vacek:
This PR add ability to load profile remotely without CLI installation. It moves profile resolution to `getProfile` method and adds check to ensure that use case exists in profile AST.

## Types of changes

## Checklist:

### DIFF
--- a/src/core/client/client.internal.ts
+++ b/src/core/client/client.internal.ts
@@ -4,11 +4,10 @@ import {
   ProfileDocumentNode,
 } from '@superfaceai/ast';
 
-import { SuperCache } from '../../lib';
+import { profileAstId, SuperCache } from '../../lib';
 import { SuperJson } from '../../schema-tools';
 import { Config } from '../config';
 import {
-  profileFileNotFoundError,
   profileNotInstalledError,
   unconfiguredProviderInPriorityError,
 } from '../errors';
@@ -34,7 +33,7 @@ export class InternalClient {
     private readonly crypto: ICrypto,
     private readonly fetchInstance: FetchInstance & Interceptable & AuthCache,
     private readonly logger?: ILogger
-  ) { }
+  ) {}
 
   /**
    * Resolves profile AST file.
@@ -52,6 +51,7 @@ export class InternalClient {
    */
   // TODO: Move to SuperfaceClientBase or move to separate file to simplify reause and testing?
   // TODO: Add logs
+  // TODO: private
   public async resolveProfileAst(
     profileId: string
   ): Promise<ProfileDocumentNode> {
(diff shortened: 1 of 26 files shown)


TheEdward162 (approved):

src/core/profile-provider/profile-provider.ts
```diff
@@ -46,7 +43,7 @@ import { resolveSecurityConfiguration } from './security';
 const DEBUG_NAMESPACE = 'profile-provider';
 
 export async function bindProfileProvider(
-  profileConfig: ProfileConfiguration,
+  profile: ProfileBase,
```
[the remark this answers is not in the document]
lukas-valenta: but leave it for now, should use IProfile from #258

src/core/client/client.internal.ts
```diff
@@ -28,11 +40,110 @@ export class InternalClient {
...
+   * File property:
+   *  - loads directly passed file
+   *  - can point only to .supr.ast.json file
+   *  - throws if file not found or not valid ProfileDocumentNode
+   * Version property:
+   *  - looks for [profileId]@[version].supr.ast.json file in superface/grid
+   *  - if not found it tries to fetch profile AST from Registry
+   * @returns ProfileDocumentNode
+   */
+  public async resolveProfileAst(
```
lukas-valenta, line 54: question: does this need to be a method on the client?

src/core/profile-provider/profile-provider.ts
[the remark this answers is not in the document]
lukas-valenta: third iteration: just use what you need, which is AST?

lukas-valenta (changes requested):

src/core/client/client.internal.ts
Jakub-Vacek: Probably not. I wasn't sure where to move it.

src/core/client/client.internal.ts
```diff
@@ -28,11 +40,110 @@ export class InternalClient {
...
+        'Reading of possible profile file failed with error %O',
+        error
+      );
+      void error;
+    }
+
+    logFunction?.('Fetching profile file from registry');
+
+    // Fallback to remote
+    // TODO: cache this somewhere (similar to CLI - .cache, config.cachePath, in memory)?
```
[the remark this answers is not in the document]
Jakub-Vacek: When I think about it I think we should leave caching AST (and looking for it) to another PR. Cache position should be probably discussed little bit more with more people.

src/core/client/client.internal.ts
Jakub-Vacek: @lukas-valenta should I move it? 😅

lukas-valenta: yeah, move it to a function near Profile

src/core/profile/resolve-profile-ast.test.ts
```diff
@@ -0,0 +1,449 @@
...
+
+// const mockSuperJsonCustomPath = new SuperJson({
+//   profiles: {
+//     test: '2.1.0',
+//   },
+//   providers: {
+//     quz: {},
+//   },
+// });
+
```
lukas-valenta, line 60:
```suggestion
```

lukas-valenta (approved):

PR #265, opened 2022-08-01
fix: missing error handling

Jakub-Vacek:
This PR fixes missing error handling when getting profile ast.

## Types of changes

## Checklist:

### DIFF
--- a/src/core/errors/errors.helpers.ts
+++ b/src/core/errors/errors.helpers.ts
@@ -508,13 +508,14 @@ export function unexpectedDigestValue(
   );
 }
 
-export function invalidResponseError(error: unknown): SDKExecutionError {
+export function invalidResponseError(
+  statusCode: number,
+  error: unknown
+): SDKExecutionError {
   if (isRegistryErrorBody(error)) {
     return new SDKBindError(
-      `Registry call failed with unexpected error: ${
-        error.detail ?? error.title
-      }`,
-      [],
+      `Registry call failed with status code: ${statusCode} and error message: ${error.title}`,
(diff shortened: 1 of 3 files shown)


src/core/errors/errors.helpers.ts
```diff
@@ -507,6 +508,24 @@ export function unexpectedDigestValue(
   );
 }
 
+export function invalidResponseError(error: unknown): SDKExecutionError {
+  if (isRegistryErrorBody(error)) {
+    return new SDKBindError(
+      `Registry call failed with unexpected error: ${
+        error.detail ?? error.title
```
freaz, line 515: suggestion(not-blocking): I would keep (show) both detail and title

src/core/registry/registry.test.ts
```diff
@@ -647,6 +648,40 @@ describe('registry', () => {
...
+
+    it('throws error on invalid status code and response body with detail', async () => {
+      const mockBody = {
+        detail: 'Test',
+        title: 'Title',
+      };
+      const mockResponse = {
+        statusCode: 400,
+        body: mockBody,
+        headers: { test: 'test' },
```
freaz, line 660: question: I don't know if relying only on body structure is right here. Server is returning header`Content-Type: application/problem+json` so the clients know what structure to expect.

freaz (approved):
Definitely missing part, only thing I am not sure about is skipping content type of the response. Otherwise looks ok.

src/core/registry/registry.test.ts
Jakub-Vacek: I can use `content-type` to decide.

PR #268, opened 2022-08-02
Metrics reporting without super.json

freaz:
## Description

Setting up metrics reporting without super.json and changing super.json to be uptional for MetricReporter instance.

Client will skip `SDKInit` event reporting if super.json is not present.

This requires changes in backend services, which are done in https://github.com/superfaceai/brain/pull/434

## Motivation and Context

We want reporting of performs even without super.json.
But SDKInit doesn't need to be send as it doesn't have any value.

## Types of changes

## Checklist:


Jakub-Vacek (approved):
Looks ok 👌

lukas-valenta (approved):

PR #276, opened 2022-08-08
feat: Remove parser

lukas-valenta:
## Description
Removes the parser dependency completely

## Motivation and Context
Parser is too big of a dependency and we decided it would be better to remove it now altogether and perhaps return it as an optional dependency later

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:


src/core/profile-provider/profile-provider.ts
```diff
@@ -80,8 +78,6 @@ export type BindConfiguration = {
 
 export class ProfileProvider {
   private profileId: string;
-  private scope: string | undefined;
```
Jakub-Vacek, line 83: Nice 👍

src/core/profile-provider/profile-provider.ts
```diff
@@ -191,26 +180,13 @@ export class ProfileProvider {
           mapVariant !== undefined
             ? `${profileId}.${providerName}.${mapVariant}@${version}`
             : `${profileId}.${providerName}@${version}`;
-        const mapSource = await fetchMapSource(
+        mapAst = await fetchMapAST(
```
Jakub-Vacek, line 183: Not sure if this fallback makes sense now. It was supposed to catch problems with AST version mismatch between brain and sdk used in client app but now we are always getting the AST and I think that if we get not valid AST from bind we will also get invalid AST from direct call so I wouldn't do this call at all. 🤔

src/core/profile-provider/profile-provider.test.ts
```diff
@@ -546,58 +544,6 @@ describe('profile provider', () => {
         expect(result).toMatchObject(expectedBoundProfileProvider);
       });
 
-      it('returns new BoundProfileProvider get map source affter failed validation', async () => {
```
Jakub-Vacek, line 549: This should now test if we fallback to fetching of map AST, I think. 🧐

Jakub-Vacek:
Looks ok, I am just not sure about the fallback to map AST fetch.

src/core/profile-provider/profile-provider.ts
freaz: Jakub is right, the AST will be the same. 

We will need either resume work on registry serving any [ast version](https://www.notion.so/superface/Registry-serving-any-AST-version-9bd743eae9cb4875ae342ea86b5fc548?v=0f4018588dc0418fbba106531b48fef5) or hope it will be fast transition to comscript.

yarn.lock
```diff
@@ -4638,9 +4633,9 @@ typescript@4.3.5:
   integrity sha512-DqQgihaQ9cUrskJo9kIyW/+g0Vxsk8cDtZ52a3NGh0YNTfpUSArXSohyUGnvbPazEPLu398C0UxmKSOrPumUzA==
 
 typescript@^4:
-  version "4.4.3"
-  resolved "https://registry.yarnpkg.com/typescript/-/typescript-4.4.3.tgz#bdc5407caa2b109efd4f82fe130656f977a29324"
-  integrity sha512-4xfscpisVgqqDfPaJo5vkd+Qd/ItkoagnHpufr+i2QCHBsNYp+G7UAoyFl8aPtx879u38wPV65rZ8qbGZijalA==
+  version "4.7.4"
```
freaz, line 4636: nitpick: Updated TS is surprising here.

package.json
```diff
@@ -46,6 +46,7 @@
     "watch": "yarn build --watch"
   },
   "devDependencies": {
+    "@superfaceai/parser": "^1.2.0",
```
freaz, line 49: question: is `^` intentional? I know it is not production dependency now, so should be fine. But asking to be sure.

freaz (commented):
Looks fine, left few comments about dependencies, but most important is fallback to AST which probably isn't needed.

package.json
Jakub-Vacek: I think it is intentional- we use parser in tests

freaz: I know to create profileId using ProfileId and ProfileVersion. But I was pointing added `^`, but dev dependencies will use yarn.lock so it is fine.

yarn.lock
lukas-valenta: :thinking: I think this is because parser depended on different version?

TheEdward162 (approved):
Is ok barring the issues @Jakub-Vacek opened

Jakub-Vacek (approved):

PR #290, opened 2022-10-05
Fix for resolving accept header

### LINKED ISSUE #264: [BUG] OneSDK overrides the custom `accept` header
Given the following http request:

```hcl
  http GET "/example" {
    request {
      headers {
        Accept = "application/json"
      }
    }
    response 200 "application/json" {
      // handle correct result
    }
    response {
      // handle everything else
    }
  }

```

## Expected Behavior

The request should be sent with `accept: application/json`.

## Current Behavior

I see the following with `DEBUG=superface:http:sensitive` env:

In case I name the header in the map as `accept` (lowercase a), I don't see it in the debug output at all.

## Possible Solution

Note that this _could_ be an issue with debug output, I didn't test what header actually reaches the endpoint in the end.

The current implementation of the interpreter [collects expected response types](https://github.com/superfaceai/one-sdk-js/blob/0bbb6f74b328294a98041a50e756659666ae26b4/src/core/interpreter/map-interpreter.ts#L328-L337) and passes them to the request. If any of the response handlers is missing a response type, `*/*` is set instead. It is likely that somewhere down the line, the passed-in `accept` parameter overrides the user-provider header.

## Your Environment

* Version used `@superfaceai/one-sdk` `1.5.0`
* Environment name and version `v16.15.0`
* Link to your project https://github.com/superfaceai/station

freaz:
## Description

Do not set `accept` header when defined in [HTTP request headers](https://spec.superface.dev/latest/map-spec.html#HTTPHeaders). I wanted to normalize headers and work with them, but that could lead to unexpected behavior and better will be to allow set as many headers and in any case. It will delegate need to normalize to client and server.

One thing I am not happy about, is how the logic is split in `map-interpreter` and `filters`. 

## Motivation and Context

Fixes #264

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- Do not set `accept` header when defined in [HTTP request headers](https://spec.superface.dev/latest/map-spec.html#HTTPHeaders) - [#264](https://github.com/superfaceai/one-sdk-js/issues/264)
 
 ## [2.0.0] - 2022-08-15
 ### Added
--- a/src/core/interpreter/http/filters.test.ts
+++ b/src/core/interpreter/http/filters.test.ts
@@ -360,5 +360,22 @@ describe('HTTP Filters', () => {
         })
       );
     });
+
+    it("doesn't set accept header if already exists", async () => {
+      const result = await headersFilter({
+        parameters: {
+          url: '/test',
+          method: '',
+          baseUrl: 'https://example.com',
+          headers: {
+            aCcEpT: 'application/json',
+          },
+          accept: 'text/plain',
+          contentType: JSON_CONTENT,
+        },
+      });
+
+      expect(result.request?.headers?.accept).toBe(undefined);
+    })
   });
 });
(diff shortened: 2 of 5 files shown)


TheEdward162 (approved):
I think the split is fine architecturally because the part in map-interpreter is responsible for collecting the information from the map (which just happens to be in a format that is the same as the header value), while the filtering is something that is specific to the HTTP client itself?

freaz:
I don't think it is entirely wrong, but I would just like to see headers resolution in map-interpreter instead of filters. 

But more importantly, when I looked at PR without focus on Accept header, I realized we work with case-sensitive keys for header names, and use of [content-type](https://github.com/superfaceai/one-sdk-js/blob/dev/src/core/interpreter/http/filters.ts#L265) and [Content-Type](https://github.com/superfaceai/one-sdk-js/blob/dev/src/core/interpreter/http/filters.ts#L274) can lead to interesting bugs.

freaz:
I would merge this one, and apply similar approach for `content-type` and `user-agent` in `headersFilter`.

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Do not set `accept` header when defined in [HTTP request headers](https://spec.superface.dev/latest/map-spec.html#HTTPHeaders)
```
jnv, line 11: nitpick (non-blocking): Consider linking to #264 from the changelog for reference

PR #293, opened 2022-10-11
Shape HeadersInit from FetchParameters

freaz:
## Description

Added function to format Record to array of tuples which is expected in `HeadersInit` and support duplicate headers.

The original test is indented, so it looks long. So, check it with [hidden whitespaces](https://github.com/superfaceai/one-sdk-js/pull/293/files?diff=split&w=1).

## Motivation and Context

TypeScript compiler was complaining about the shape of `headersInit` data.

## Types of changes

## Checklist:

### DIFF
--- a/src/node/fetch/fetch.node.test.ts
+++ b/src/node/fetch/fetch.node.test.ts
@@ -15,225 +15,269 @@ const timers = new MockTimers();
 
 type ForEachCallbackFunction = (value?: string, type?: string) => void;
 
-describe('fetch', () => {
-  beforeEach(async () => {
-    await mockServer.start();
-  });
-
-  afterEach(async () => {
-    await mockServer.stop();
-    jest.resetAllMocks();
-    jest.resetModules();
-  });
-
-  describe('timeout', () => {
-    it('timeouts on network timeout', async () => {
-      // we want to use actuall fetch implementation
-      mocked(fetch).mockImplementation(jest.requireActual('cross-fetch').fetch);
-
-      await mockServer.forGet('/test').thenTimeout();
-      const realTimers = new NodeTimers();
-      const nodeFetch = new NodeFetch(realTimers);
-
-      await expect(
-        nodeFetch.fetch(`${mockServer.url}/test`, { method: 'GET', timeout: 2000 })
-      ).rejects.toEqual(new NetworkFetchError('timeout'));
-    });
-
-    it('rejects on rejected connection', async () => {
-      // we want to use actuall fetch implementation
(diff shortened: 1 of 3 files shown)


src/node/fetch/fetch.node.ts
```diff
@@ -239,4 +234,22 @@ export class NodeFetch implements IFetch, Interceptable, AuthCache {
...
+
+    const headers: [string, string][] = [];
+
+    Object.entries(data).forEach(([key, value]) => {
+      if (Array.isArray(value)) {
+        value.forEach((val) => headers.push([key, val]));
+      } else {
+        headers.push([key, value]);
+      }
+    });
```
jnv, line 251: thought: Would using `flatMap` make this more readable?

```suggestion
    headers = Object.entries(data).flatMap(([key, value]) => {
      if (Array.isArray(value)) {
        return value.map((nestedVal) => [key, nestedVal])
      }
      return [[key, value]]
    }
```

But it's probably less performant (but so is `forEach`).

freaz: Thank you, but I intentionally implemented this way.

src/node/fetch/fetch.node.test.ts
```diff
@@ -15,225 +15,292 @@ const timers = new MockTimers();
...
+      const binaryContentTypes = [
+        'application/octet-stream',
+        'audio/mp3',
+        'audio/wav',
+        'audio/wav;rate=8000',
+        'video/mp4',
+        'image/jpeg',
+      ];
+
+      for (const contentType of binaryContentTypes) {
```
TheEdward162, line 215: question: any idea why this doesn't use https://jestjs.io/docs/api#testeachtablename-fn-timeout

src/node/fetch/fetch.node.ts
TheEdward162: I think at that point I would find a classic `for .. of` loop more readable, but I think this code is short enough to be fine either way.

TheEdward162 (approved):

src/node/fetch/fetch.node.test.ts
freaz: Don't know. But I just rewrote when I changed `content-type` handling in https://github.com/superfaceai/one-sdk-js/blob/dev/src/core/interpreter/http/filters.ts#L259

freaz: Now (when I ran the test) I know why. `beforeEach` can't be used inside `test.each`.

PR #295, opened 2022-10-16
streaming map interpreter

TheEdward162:
## Description
Redesign of map interpreter using async generators. For now, this will only change internal details and should not change any public behavior.

## Motivation and Context
This will open the possibility of adding a `yield` keyword to Comlink and allow us to return partial results - turning `perform` (or `performStream`) into a generator.

## Types of changes

## Checklist:


TheEdward162:
I've also run station tests against this by using the https://github.com/superfaceai/station/tree/chore/onesdk_2_1_0 branch + `yalc add` and ignoring `mock.test.ts` files with `yarn test --testPathIgnorePatterns 'mock.test.ts' --runInBand`

TheEdward162:
Also - it is not true that this PR doesn't change public API surface. I have added `UnexpectedError` and `SDKExecutionError` into the return signature of `MapInterpreter` and `BoundProfileProvider` because it was always supposed to be there - so in that sense it is a bugfix rather than a minor change.

src/core/interpreter/map-interpreter.ts
```diff
@@ -85,770 +72,1318 @@ export interface MapParameters<
...
+              node = operations[step.value.what.operation];
+              if (node === undefined) {
+                return err(
+                  new MapASTError(
+                    `Operation not found: ${step.value.what.operation}`,
+                    { node: current.node, ast }
+                  )
+                );
+              }
+            }
```
freaz, line 1241: question: handling operation calls seems to be disconnected from rest of AST walking. Is it because we need to jump in AST tree to operation?

freaz (commented):
I got through approx. half of it. But started to comment on `MapInterpreter`.

src/core/interpreter/map-interpreter.ts
TheEdward162: Yes, the options I considered were to either pass `operations` record into the visitor or to have the visitor communicate that it wants to call an operation.

I opted for the second option as it centralizes error handling if the operation is not found and keeps the convention that visitors do not have access to the entire AST but rather only the relevant node.

src/core/interpreter/map-interpreter.ts
```diff
@@ -85,770 +72,1318 @@ export interface MapParameters<
...
     }
 
-    return stack;
+    // create a visitor of the root node and put it on the stack
+    const nodeStack: NodeVisitor<MapASTNode, unknown>[] = [
+      this.createVisitor(
+        entry,
+        {
+          input: this.parameters.input ?? {},
+          parameters: this.parameters.parameters ?? {},
```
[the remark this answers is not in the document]
TheEdward162: The convention I wanted to have here is that a child is free to edit its stack - it should be caller-saved. The fact that the parent then uses the stack directly is the parent's choice.

Either way, a good point that this allows children to change input and integration parameter values and the parent will accept that. This wasn't possible before as input and integration parameters were injected into the frame in `get variables()`.

I will write a stack mutation test and then fix it by cloning the stack before passing it to a child.

TheEdward162:
rebased on top of `dev`

TheEdward162:
rebased on `dev`

PR #296, opened 2022-10-28
Binary data implementation

lukas-valenta:
## Description
In this pull request is introduced new "primitive" type `BinaryData`. 

```ts
export interface IBinaryData {
  peek(size?: number): Promise<Buffer | undefined>;
  getAllData(): Promise<Buffer>;
  chunkBy(chunkSize: number): AsyncIterable<Buffer>;
  toStream(): Readable;
}
```

If `BinaryData` instance gets to be used as body or value in `FormData`, it is streamed.
If `BinaryData` should be returned as outcome, an `UnexpectedError` is returned. To solve it, call `getAllData()` to read stream into `Buffer`. Later streamed results will be added.

### Limitations

**1. Asynchronous operations on BinaryData**
Current limitation for use in Comlink Map is that `peek`, `read` and `getAllData` methods are asynchronous. So, they can't be directly chained with function on [Buffer](https://nodejs.org/dist/latest-v18.x/docs/api/buffer.html#class-buffer). Most likely `toString()`.

To overcome this limitation, It must be first [set into variable](https://spec.superface.dev/latest/map-spec.html#sec-Set-Variables), so the interpreter resolves the Promise:
```
map Example {
  firstBytes = input.binary.peek(10)
  data = input.binary.read(10)
  remainingData = input.binary.getAllData()

  map result {
    firstBytes = firstBytes
    data = data
   remainingData = remainingData
  }
}
```

**Opaque data type**
In profile, the type must be omitted for now. Later we may add it to Comlink Profile as [new Primitive](https://spec.superface.dev/latest/profile-spec.html#sec-Primitive-types).

```
name = "example"
version = "1.0.0"

usecase Example {
  input {
    binary
  }
}

```

### Comlink Map examples

Send file in body:
```
map SendFile {
  http POST "/resource" {
    request "application/octet-stream" {
      body = input.file
    }

    response 200 "text/plain" {
      map result body
    }
  }
}
```

Send file as FormData field:
```
map SendFile {
  http POST "/resource" {
    request "multipart/form-data" {
      body {
        field = input.field
        file1 = input.file1
        file2 = input.file2
      }
    }

    response 200 "text/plain" {
      map result body
    }
  }
}
```

Send file by chunks:
```
profile = "example@1.0"
provider = "localhost"

map ChunkedUpload {
  uploadResults = call foreach(chunk of input.file.chunkBy(10)) Upload(chunk = chunk)

  return map result uploaded
}

operation Upload {
  http POST "/upload/{args.uploadId}" {
    request "application/octet-stream" {
      body = args.chunk
    }

    response 200 "text/plain" {
      return true
    }

    response 500 "text/plain" {
      return false
    }
  }
}
```

Read first bytes of file and then send it in body:
```
map SendFile {
  firstTen = input.file.peek(10)

  contentType = firstTen.toString('ascii').toLowerCase().includes('pdf') ? 'application/pdf' : 'application/octet-stream'

  http POST "/resource" {
    request {
	  headers {
        'content-type' = contentType
      }
    
      body = input.file
    }

    response 200 "text/plain" {
      map result body
    }
  }
}
```

### OneSDK use example

From user perspective it is used as followed:

```typescript
import { SuperfaceClient, BinaryData } from '@superfaceai/one-sdk';

const client = new SuperfaceClient();
const profile = await client.getProfile('example');

// for files
const result = await profile.getUseCase('Example').perform({
  binary: BinaryData.fromPath(pathToFile, { mimetype: 'text/plain', filename: 'example.txt' });
});

// or for Readable streams
const stream = getReadableStream();
const result = await profile.getUseCase('Example').perform({
  binary: BinaryData.fromStream(stream);
});
```

## Motivation and Context

To allow the passing of binary data with capabilities like async loading, streaming and chunking.

## Types of changes

## Checklist:


src/node/filesystem/binary.node.ts
```diff
@@ -0,0 +1,122 @@
...
+  IEncodable,
+  IInitializable,
+} from '../../interfaces';
+import { UnexpectedError } from '../../lib';
+
+export class ChunkedFile implements IChunked {
+  constructor(private binaryFile: BinaryFile) {}
+
+  public async getChunk(): Promise<string> {
+    return this.binaryFile.getNextChunk();
```
TheEdward162, line 18: issue: this seems really brittle. What if attempt something like `file.chunkBy(10).skip(1)`? This will effectively skip the last chunk and not the first one. In addition, since the state is stored in the original file it will also break if I attempt to chunk the same file twice.

src/node/filesystem/binary.node.ts
```diff
@@ -0,0 +1,341 @@
...
+    let stream: Readable;
+    // We need to create a stream the old fashioned way for Node < 16.11.0
+    if (typeof this.handle.createReadStream !== 'function') {
+      stream = createReadStream(this.path);
+    } else {
+      stream = this.handle.createReadStream();
+    }
+
+    this.stream = new ReadableClone(stream);
+    this.streamReader = new StreamReader(new ReadableClone(stream));
```
TheEdward162, line 153: question: I don't quite understand the point of this cloning. What is it good for?

src/node/filesystem/binary.node.ts
```diff
@@ -0,0 +1,341 @@
...
+    return chunk;
+  }
+}
+
+export class ReadableClone extends Readable {
+  constructor(private stream: Readable, options?: ReadableOptions) {
+    super(options);
+
+    this.stream.on('data', (chunk) => {
+      this.push(chunk);
```
TheEdward162, line 93: issue: I'm not sure about this, but I think this can be a big performance problems. The NodeJS docs contain a code sample which I think can be used here almost identically and it contains pausing mechanism, but that might interfere with point of this class.

https://nodejs.org/api/stream.html#readablepushchunk-encoding

TheEdward162: Ideally we would want something like this, but that's a web API - experimental and not the one we are using right now in Node https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/tee

src/node/filesystem/binary.node.test.ts
```diff
@@ -0,0 +1,293 @@
...
+      streamContainer = new StreamContainer(new MockStream());
+    });
+
+    it('creates ReadableClone for data stream', async () => {
+      expect((streamContainer as any).stream).toBeInstanceOf(ReadableClone);
+    });
+
+    it('passes ReadableClone to StreamReader', async () => {
+      expect((streamContainer as any).streamReader.stream).toBeInstanceOf(ReadableClone);
+    });
```
TheEdward162, line 146: question: should we really be testing for this? Seems like a very deep implementation detail to me.

src/node/filesystem/binary.node.ts
```diff
@@ -0,0 +1,341 @@
...
+  private stream: Readable;
+  private streamReader: StreamReader;
+
+  constructor(stream: Readable) {
+    this.stream = new ReadableClone(stream);
+    this.streamReader = new StreamReader(new ReadableClone(stream));
+  }
+
+  public read(size?: number): Promise<Buffer | undefined> {
+    return this.streamReader?.read(size);
```
TheEdward162, line 189: nitpick: probably doesn't need a `?` anymore?

src/node/filesystem/binary.node.ts
```diff
@@ -0,0 +1,341 @@
...
+        },
+      }),
+    };
+  }
+
+  public async getAllData(): Promise<Buffer> {
+    let size = this.buffer.length;
+
+    while (this.buffer.length >= size) {
+      size += 1000; // TODO constant
```
TheEdward162, line 313: thoughts: ideally a configurable constant, and probably should be bigger like 16 or even 64kB to make sense for bigger files. Although we might have bottlenecks in other places anyway.

src/node/filesystem/binary.node.ts
TheEdward162: Oh, and that code is *not* trivial https://github.com/nodejs/node/blob/f9227ffc4b47da1ccc0b621001423f3221eca27a/lib/internal/webstreams/readablestream.js#L1451

freaz: Thats what I wrote you about yesterday. And the code is comming from nodejs do documemtstion. Right now we need to either clone streamed data or getAllData to get data feom stream/file before sending it in request.

Anotger problem is, that if user consumes data from BinaryData and still passes BinaryData as payload it will be empty.

src/node/filesystem/binary.node.ts
freaz: You are right. I left this for now to hand over rest of changes. But thanks for suggestions.

src/core/interpreter/sandbox/sandbox.ts
```diff
@@ -11,11 +14,18 @@ function vm2ExtraArrayKeysFixup<T>(value: T): T {
...
-  if (value === null) {
+  if (value === null || value === undefined) {
     return value;
   }
 
-  if (Buffer.isBuffer(value) || value instanceof ArrayBuffer) {
+  if (
+    Buffer.isBuffer(value) ||
+    value instanceof ArrayBuffer ||
+    isBinaryData(value) || // maybe not needed?
```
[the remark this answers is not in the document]
freaz: Do you have idea how to get rid if it? This was the reason, why I jad comment it doesn’t work somewhere else.  And was surprised by this.

src/core/interpreter/map-interpreter.ts
```diff
@@ -531,10 +546,14 @@ export class MapInterpreter<TInput extends NonPrimitive | undefined>
...
-        ast: this.ast,
-      });
+      if (e instanceof Error) {
+        throw new JessieError('Error in Jessie script', e, {
+          node,
+          ast: this.ast,
+        });  
+      }
+
+      throw e;
```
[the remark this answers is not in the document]
freaz: That’s the messed up type in errors again. And I forgot to return it to original state and just suppress TSC as in other cases because I plan to look at error handling as a whole.

src/interfaces/usecase.ts
```diff
@@ -24,3 +24,68 @@ export interface IUseCase {
...
+  chunkBy(offset: number): Iterable<IChunked>;
+  peek(size: number): string;
+}
+
+export interface IInitializable {
+  initialize(): Promise<void>;
+}
+
+export interface IDestructible {
+  destroy(): Promise<void>;
```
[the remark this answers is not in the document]
freaz: I completely forgot about this. I will have a look if and how to add it.

src/core/interpreter/map-interpreter.test.ts
```diff
@@ -1419,6 +1428,450 @@ describe('MapInterpreter', () => {
...
+    expect(result.isOk()).toBe(true);
+    const firstBytes = (result.unwrap() as { firstBytes: Buffer }).firstBytes.toString('utf8');
+    const items = (result.unwrap() as { items: Buffer }).items;
+    const expected = await readFile(filePath);
+
+    expect(firstBytes).toStrictEqual(expected.subarray(0, 10).toString('utf8'));
+    expect(items).toStrictEqual(expected);
+  });
+
+  it('should resolve BinarData as Buffer if mapped as result', async () => {
```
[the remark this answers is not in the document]
freaz: Isn't it why we have #295? So data can be yielded from map?

src/core/interpreter/map-interpreter.test.ts
```diff
@@ -1419,6 +1428,450 @@ describe('MapInterpreter', () => {
...
+    expect(response).toStrictEqual({ ok: true });
+  });
+
+  it('should send file as a stream in body', async () => {
+    const filePath = path.resolve(process.cwd(), 'fixtures', 'binary.txt');
+    const file = BinaryData.fromPath(filePath);
+    const expected = await readFile(filePath, 'utf8');
+    await mockServer.forPost('/test').thenCallback(async req => {
+      const body = await req.body.getText();
+      expect(body).toStrictEqual(expected);
```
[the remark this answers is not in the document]
freaz: Do you mean the `toStrictEqual` or the fact that the assertion is in callback? If later one, then both of us are in wonderland. :D

src/node/filesystem/binary.node.ts
freaz: Yes, you are right. I left it as to-do, because wanted to hand over the rest.

src/node/filesystem/binary.node.ts
freaz: You are right. It will be oversight leftover from quick prototyping. Thanks

src/core/interpreter/sandbox/sandbox.ts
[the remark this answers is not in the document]
freaz: Yeah, it is weird, I got bitten by it and then found this gem.

src/node/filesystem/binary.node.ts
freaz: All streams handling was changes. Resolving as outdated.

src/node/filesystem/binary.node.ts
freaz: Added chunkSize argument with 16000 (16kB) as default

src/node/filesystem/binary.node.test.ts
```diff
@@ -11,23 +11,34 @@ const fixturePath = joinPath('fixtures', 'binary.txt');
...
   private max: number;
+  private readSize: number;
   private index: number;
 
-  constructor(max = 10, options?: ReadableOptions) {
+  constructor(max = 10, readSize = 1, options?: ReadableOptions) {
     super(options);
     this.max = max;
-    this.index = 1;
+    this.readSize = readSize;
```
TheEdward162, line 20: praise: I really like how the readSize here allows the tests to cover both "remaining in internal buffer, stream still going" and "remaining in internal buffer, stream done"

src/node/filesystem/binary.node.ts
```diff
@@ -28,7 +29,7 @@ export class StreamReader {
   private endCallback: (this: () => void) => void;
 
   constructor(stream: NodeJS.ReadableStream) {
-    this.buffer = Buffer.from([]);
+    this.buffer = Buffer.alloc(0);
```
TheEdward162, line 32: praise: noice, this feels cleaner somehow

src/node/filesystem/binary.node.ts
```diff
@@ -172,43 +178,29 @@ export class FileContainer implements IDataContainer, IBinaryFileMeta, IInitiali
...
-
-    return this.stream;
+    return this.streamReader.toStream();
   }
 }
 
 export class StreamContainer implements IDataContainer {
   private stream: NodeJS.ReadableStream;
-  private streamReader: StreamReader | undefined;
+  private streamReader: StreamReader;
```
TheEdward162, line 191: question: Do we still need to explicitly hold the `stream` if we have it in `streamReader` and `toStream` is implemented through it as well?

src/interfaces/usecase.ts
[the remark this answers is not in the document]
freaz: Since we are using it for interfaces which are implemented by classes, Symbols would be hard to use.

src/node/filesystem/binary.node.test.ts
freaz: Outdated

TheEdward162 (approved):

PR #297, opened 2022-11-07
fix(interpreter): `/` url will not error as absolute

TheEdward162:
## Description
The following map was returning an error about the url not being relative, even though it is considered relative to the base url and should work.
```
http GET '/' {}
```

## Motivation and Context
Bug reported through Slack

## Types of changes

## Checklist:


freaz (approved):
Thanks for fix and proper test coverage.

PR #304, opened 2022-12-09
Allow setting binary data metadata

freaz:
## Description

- adds `name` and `mimetype` to `fromStream` factory function
- adds setters on `BinaryData` for `name` and `mimetype`

### Example use in map

```
map Example {
  // need to do comlink variable assign to be able to use Script expression
  tmp = (input.file.name = input.filename)
  tmp = (input.file.mimetype = input.mimetype)
  
  http POST "/" {
    request "multipart/form-data" {
      body = {
        file: input.file
      }
    }
  
    response 200 {
      map result body
    }
  }
}
```

## Motivation and Context

Comlink author may need to set mimetype or name from within the map.

## Types of changes

## Checklist:

### DIFF
--- a/src/node/filesystem/binary.node.ts
+++ b/src/node/filesystem/binary.node.ts
@@ -133,13 +133,20 @@ export class StreamReader {
 export class FileContainer implements IDataContainer, IBinaryDataMeta, IInitializable, IDestructible {
   private handle: FileHandle | undefined;
   private streamReader: StreamReader | undefined;
-  public size = Infinity;
-  public name: string | undefined;
-  public mimetype: string | undefined;
+  private _size = Infinity;
 
-  constructor(public path: string, options: { filename?: string, mimetype?: string } = {}) {
-    this.mimetype = options.mimetype;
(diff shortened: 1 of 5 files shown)


src/node/filesystem/binary.node.ts
```diff
@@ -133,13 +133,21 @@ export class StreamReader {
 export class FileContainer implements IDataContainer, IBinaryDataMeta, IInitializable, IDestructible {
   private handle: FileHandle | undefined;
   private streamReader: StreamReader | undefined;
-  public size = Infinity;
-  public name: string | undefined;
-  public mimetype: string | undefined;
+  public _size = Infinity;
+  public _mimetype: string | undefined;
```
TheEdward162, line 137: question: Can these be private instead?

TheEdward162 (approved):
That map hack is disgusting but that is not a problem of this PR :D

src/node/filesystem/binary.node.ts
freaz: 🤦 they must by private. Thanks

Issue #311, opened 2022-12-21
[BUG] nullable enum input validation


Jakub-Vacek:
Provide a general summary of the issue in the Title above

## Expected Behavior
Perform should pass

## Current Behavior
Profile defines input field units enum {C, F, K}  (optional enum, can be null) but when you specify units as null during perform call:

```
await profile
    .getUseCase('GetCurrentWeatherInCity')
    .perform({
      city: null,
      units: null
    }, {
      provider: 'wttr-in'
    });
```

SDK throws error:

```
InputValidationError {
  kind: 'InputValidationError',
  message: 'Input validation failed:\nPath: input.units\nError: Invalid enum value: null',
  errors: [ { kind: 'enumValue', context: [Object] } ],
  name: 'InputValidationError'
}

```

## Possible Solution
Not obligatory, but suggest a fix/reason for the bug

## Steps to Reproduce
Provide a link to a live example, or an unambiguous set of steps to reproduce this bug. Include code to reproduce, if relevant
1. Use profile with nullable enum as input
2. Run perform with enum value set to null

## Your Environment
Include as many relevant details about the environment you experienced the bug in. Preferably include:
* 2.1.0, discovered on 2.2.0-rc.3

PR #312, opened 2022-12-21
Fix/nullable enum validation

Jakub-Vacek:
## Description
Fixes: https://github.com/superfaceai/one-sdk-js/issues/311

## Types of changes

## Checklist:

### DIFF
--- a/src/core/interpreter/profile-parameter-validator.ts
+++ b/src/core/interpreter/profile-parameter-validator.ts
@@ -232,6 +232,10 @@ export class ProfileParameterValidator implements ProfileVisitor {
         return [true];
       }
 
+      if (input === null) {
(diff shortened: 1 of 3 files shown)


src/core/interpreter/profile-parameter-validator.ts
```diff
@@ -232,6 +232,10 @@ export class ProfileParameterValidator implements ProfileVisitor {
         return [true];
       }
 
+      if (input === null) {
```
TheEdward162, line 235: nitpick: I mean you could have merged it with the above if :D

TheEdward162 (approved):
Nice. I think other things might need the same treatment (arrays, for example) as well. I'll try to fix those.

PR #314, opened 2022-12-21
Shared: Introduce `None`  type

freaz:
## Description

Adds `None` as possible `Variables` value.

## Motivation and Context

Incorrect handling of `null` and `undefined` lead to confusion by end users, and we internally introduced bugs as TypeScript types didn't allow `null` as a possible value.

Updated specification: https://github.com/superfaceai/spec/pull/47

[Decision Record](https://www.notion.so/superface/None-type-fc921ac39d764a88ad5ba3b1cc3d148a) _internal_

### Context

When we were implementing Binary Data, we discovered a major flaw in types and how they interact with required/optional fields and nullable values. 

We now have only JavaScript environment.

### Decision

`None` is defined as union of `undefined` and `null`.

It will be added as possible `Variables` value.

`None` CANNOT be used as primitive type, it is implicit type for `nullable` values.

We will add `NoneLiteral` to support `undefined` and `null` values in Examples.

By default, fields are `Optional`.

Required field MUST fail validation only if the field in the object isn’t defined at all. So, it doesn’t matter whether the value is `None` or any other Model. Behaviour similar to `Object.hasOwn()` in JavaScript.

By default, value is `Optional`.

`Required` value CANNOT be `None`.

#### Example

```rust
usecase Example {
  input {
    field1 string // optional value can be String or None, in JS: 'value', undefined, null
    field2 string! // required value can be String, in JS: 'value'
  }

  example Success {
    input {
      field1 = None
			field2 = 'value'
    }
  }
}
```

If we surfaced types for profile’s input, field1 would be typed as `Optional<string>`, which can be represented also as `string | None` and fully expanded as `string | null | undefined`.

For field2, type is simply `string`.

### Consequences

Clearly defined behaviour of absence of the value for user, as well internal handling of `undefined` and `null` values.

## Types of changes

## Checklist:

### DIFF
--- a/src/lib/variables/variables.ts
+++ b/src/lib/variables/variables.ts
@@ -61,9 +61,7 @@ export function isEmptyRecord(
   return isNonPrimitive(input) && Object.keys(input).length === 0;
 }
 
-export function assertIsVariables(
-  input: unknown
-): asserts input is Variables {
+export function assertIsVariables(input: unknown): asserts input is Variables {
   if (!isVariables(input)) {
     throw new UnexpectedError(`Invalid result type: ${typeof input}`);
   }
--- a/src/lib/variables/variables.ts
+++ b/src/lib/variables/variables.ts
@@ -1,7 +1,6 @@
 import type { IBinaryData } from '../../interfaces';
 import { isBinaryData } from '../../interfaces';
 import { UnexpectedError } from '../error';
-import { indexRecord } from '../object';
 
 export type None = undefined | null;
 export type Primitive =
@@ -110,28 +109,6 @@ export function mergeVariables(
       result[key] = right[key];
     }
   }
-
-  return result;
-}
-
-export function getValue(
-  variables: NonPrimitive,
-  key: string[]
-): Variables | undefined {
-  if (variables === undefined) {
-    return undefined;
-  }
-
-  let result;
-  try {
-    result = indexRecord(variables, key);
-  } catch (_err) {
-    // return undefined on error to preserve the original behavior of this function
-    return undefined;
-  }
-
-  // sanity check, but if the input `variables` is correct then the result will be also
-  assertIsVariables(result);
 
   return result;
 }
--- a/src/lib/variables/variables.ts
+++ b/src/lib/variables/variables.ts
@@ -44,6 +44,7 @@ export function isPrimitive(input: unknown): input is Primitive {
 export function isNonPrimitive(input: unknown): input is NonPrimitive {
   return (
     typeof input === 'object' &&
+    input !== null &&
     !Array.isArray(input) &&
     !isBinaryData(input) &&
     !Buffer.isBuffer(input) &&
--- a/src/lib/variables/variables.ts
+++ b/src/lib/variables/variables.ts
@@ -133,7 +133,7 @@ export function variableToString(variable: Variables): string {
 }
 
 /**
- * Stringifies a Record of variables. `undefined` values are removed.
+ * Stringifies a Record of variables. `None` values are removed.
  */
 export function variablesToStrings(
   variables: NonPrimitive
--- a/src/lib/variables/variables.ts
+++ b/src/lib/variables/variables.ts
@@ -3,16 +3,17 @@ import { isBinaryData } from '../../interfaces';
 import { UnexpectedError } from '../error';
 import { indexRecord } from '../object';
 
-// Arrays should be considered opaque value and therefore act as a primitive, same with
+export type None = undefined | null;
 export type Primitive =
   | string
   | boolean
   | number
-  | unknown[]
+  | unknown[] // Arrays should be considered opaque value and therefore act as a primitive, same with
+  | None
   | IBinaryData
   | Buffer;
 export type NonPrimitive = {
-  [key: string]: Primitive | NonPrimitive | undefined;
+  [key: string]: Primitive | NonPrimitive;
 };
 export type Variables = Primitive | NonPrimitive;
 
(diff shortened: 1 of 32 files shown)


src/lib/variables/variables.ts
```diff
@@ -25,72 +26,70 @@ export function isClassInstance(input: unknown): boolean {
...
-export function isNonPrimitive(input: Variables): input is NonPrimitive {
+export function isNonPrimitive(input: unknown): input is NonPrimitive {
   return (
     typeof input === 'object' &&
     !Array.isArray(input) &&
-    !isClassInstance(input) &&
+    !isBinaryData(input) &&
     !Buffer.isBuffer(input) &&
-    !isBinaryData(input)
+    !isClassInstance(input)
```
TheEdward162, line 50: issue: I believe this should also be checking for `input !== null`

src/lib/variables/variables.ts
```diff
@@ -112,12 +111,12 @@ export const mergeVariables = (
...
-export const getValue = (
-  variables: NonPrimitive | undefined,
+export function getValue(
+  variables: NonPrimitive,
   key: string[]
-): Variables | undefined => {
+): Variables | undefined {
   if (variables === undefined) {
     return undefined;
   }
```
TheEdward162, line 122: suggestion: This should never happen (in types) - `NonPrimitive` is not allowed to be `undefined`. I think it can be removed.

TheEdward162: opinion: Also depending on where this is called maybe we should remove the try-catch below and let this function throw on errors so that we don't mix `undefined` as a valid value and `undefined` as a sentinel for error.

TheEdward162: Also also `Variables | undefined` is redundant because `Variables` already contains `None` which contains `undefined` - this is why I think we should enable the throwing behavior again - but that's a breaking change.

src/lib/variables/variables.ts
```diff
@@ -134,34 +133,42 @@ export const getValue = (
...
  * Turns a variable (both primitive and non-primitive) into a string.
  */
-export const variableToString = (variable: Variables): string => {
+export function variableToString(variable: Variables): string {
   if (typeof variable === 'string') {
     return variable;
   }
 
+  if (variable === undefined) {
+    return 'undefined'; // TODO: for both undefined and null, return 'None'?
```
TheEdward162, line 147: opinion: It makes sense to me, but it is probably a breaking change, even though I hope nobody is stringifying undefined into url parameters.

src/lib/variables/variables.ts
```diff
@@ -134,34 +133,42 @@ export const getValue = (
...
+}
 
 /**
  * Stringifies a Record of variables. `undefined` values are removed.
  */
-export const variablesToStrings = (
-  variables?: Variables
-): Record<string, string> => {
+export function variablesToStrings(
+  variables: Variables
```
TheEdward162, line 161: question: Looking at the documentation of this function now - should this even be taking `Variables` as input? Shouldn't this be taking `NonPrimitive` instead - if the input should always be a record...

If so, then the first if can be removed.

It could mean that we would need to update call sites.

src/lib/variables/variables.ts
```diff
@@ -134,34 +133,42 @@ export const getValue = (
...
+): Record<string, string> {
   const result: Record<string, string> = {};
 
-  if (variables !== undefined) {
+  if (!isNone(variables)) {
     for (const [key, value] of Object.entries(variables)) {
-      if (value !== undefined) {
+      if (!isNone(value)) {
```
TheEdward162, line 167: question: What is the purpose of removing the `undefined` from input here? Could this be one of the places where there is an actual difference between `null` and `undefined` in that null is always deliberate while `undefined` could be a sentinel for absence? I don't know how to approach this.

src/lib/variables/variables.ts
freaz: You are right, that first condition is not needed, and I will remove it.

Regarding `getValue` removing all the things you correctly pointed, it looses reason to exists. Mainly because it is used in one place only (util's `replaceParameters`). So, I would move try-catch and getting value there.

src/lib/variables/variables.ts
freaz: Indeed, it should. Thank you.

src/lib/variables/variables.ts
freaz: I thought about it as well, but when I looked how it is used, I found it in `replaceParameters` where it is concatenated to string, so the result will be the same.

But where it can cause problems are filters and http, where it is used to create headers, body, etc. So, it would send `'undefined'`, which is not desired. 

I will explicitly return `undefined`, so it is not left to knowledge of `JSON.stringify` behavior.

freaz: But that also means `variableToString`, should be returning `string | undefined` because now it is not clear it can actually return `undefined`.

freaz: And I was wrong in original comment because in filters and http is used `variablesToStrings`, which is removing `undefined`. And actually, it is removing `None` so even `null`. Which might be a problem. Even though previously it would return `'null'` which feels even worse.

src/lib/variables/variables.ts
freaz: I find this function more and more difficult to understand. Even the fact that NonPrimitve's  values are serialized using `JSON.stringify` seem incorrect.

And if I understood original behavior, then null was stringified `'null'`. Discussed on `variableToString`.

src/lib/variables/variables.ts
freaz: I started with it. Let's see if that's the right thingy

freaz:
Rebased on top of `dev`

PR #320, opened 2023-01-12
Add Node 19 to CI Tests

freaz:
## Description

Runs OneSDK on Node 19.

Found bug in [mockttp](https://github.com/httptoolkit/mockttp/issues/107), where connection header is deleted and client doesn't know how to behave. 

I choose to use `Connection: close` because with keep-alive, I would need to set also `Keep-Alive: timeout=5`. And I hope I will be able to fix mockttp eventually, so this can be reverted.

## Motivation and Context

I want to have OneSDK tested on all supported Node.js versions

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [14, 16, 18]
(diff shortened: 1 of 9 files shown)


TheEdward162 (approved):
I wonder how this will turn out with the next LTS, somebody ought to be running into the same trouble.

PR #323, opened 2023-01-13
feat: ensure None type behaves correctly

TheEdward162:
Good news: Tests now compile and all map interpreter tests pass, no less tests pass than in base. Verified by quickly cooking up a differ[0] and checking `jest --json --outputFile tests.json` on the base branch vs. this branch (`python3 differ.py tests1.json tests2.json`).

This contains two BREAKING CHANGEs:
* rejects values passed to headers and query parameters which aren't `string | string[]`, instead telling the user to stringify them themselves before use
  - Relaxed for booleans and numbers, which are stringified by the implementation.
* does not even set `input` nor `parameter` in interpreter if they are passed in as `undefined` or `null` (although with null it is an invalid use of the perform API).
  - This means the map will fail at runtime (jessie error) if you try to access `input` object when there is no input

It also fixes query parameter handling which was apparently broken across 3 places, so now we can pass string arrays to query as well. yay.

<details>
<summary>[0] Differ:</summary>

```python
import json
import sys

with open(sys.argv[1], "r") as file:
	left_json = json.load(file)

with open(sys.argv[2], "r") as file:
	right_json = json.load(file)

left = left_json["testResults"]
right = right_json["testResults"]

def find_suite(results, name):
	for (i, suite) in enumerate(results):
		if suite["name"] == name:
			return i
	
	return None

def find_test(tests, name):
	for (i, tests) in enumerate(tests):
		if tests["fullName"] == name:
			return i
	
	return None

def reverse_side(side):
	return "right" if side == "left" else "left"

def diff_tests(left, right, out, side = "left"):
	left_i = 0
	while True:
		if left_i >= len(left):
			break

		left_test = left[left_i]
		right_i = find_test(right, left_test["fullName"])

		if right_i is None:
			out.append(
				(side, "test", left_test["fullName"], left_test["status"])
			)
			del left[left_i]
			continue
	
		right_test = right[right_i]
		if left_test["status"] != right_test["status"]:
			if left_test["status"] != "passed":
				out.append(
					(side, "test", left_test["fullName"], left_test["status"])
				)
			else:
				out.append(
					(reverse_side(side), "test", right_test["fullName"], right_test["status"])
				)

		del left[left_i]
		del right[right_i]

	if len(right) > 0:
		diff_tests(right, left, out, side = reverse_side(side))

def diff_results(left, right, out, side = "left"):
	left_i = 0
	while True:
		if left_i >= len(left):
			break

		left_suite = left[left_i]
		right_i = find_suite(right, left_suite["name"])
		
		if right_i is None:
			out.append(
				(side, "suite", left_suite["name"], left_suite["status"])
			)
			del left[left_i]
			continue
		
		right_suite = right[right_i]
		if left_suite["status"] != right_suite["status"]:
			left_tests = left_suite["assertionResults"]
			right_tests = right_suite["assertionResults"]
			diff_tests(left_tests, right_tests, out)

		del left[left_i]
		del right[right_i]
	
	if len(right) > 0:
		diff_results(right, left, out, side = reverse_side(side))

out = []
diff_results(left, right, out)
for diff in out:
	print(diff)

```

</details>

### DIFF
--- a/src/core/interpreter/map-interpreter.ts
+++ b/src/core/interpreter/map-interpreter.ts
@@ -41,12 +41,14 @@ import {
   assertIsVariables,
   castToVariables,
   err,
+  fromEntriesOptional,
+  isNone,
   isNonPrimitive,
   isPrimitive,
   mergeVariables,
   ok,
   SDKExecutionError,
-  UnexpectedError,
+  UnexpectedError
 } from '../../lib';
 import type { IServiceSelector } from '../services';
 import type { MapInterpreterExternalHandler } from './external-handler';
@@ -246,7 +248,8 @@ abstract class NodeVisitor<N extends MapASTNode, V = undefined>
       this.visitGenerator = this.visit();
     }
 
-    // here we check child identifier as a sanity check
+    // here we check child identifier as a sanity check - this helps to ensure that we don't receive a child result
+    // that we didn't previously request using `explore` result
     const actualChildIdentifier = args[0]?.childIdentifier;
     if (this.expectedChildIdentifier !== actualChildIdentifier) {
       const expected = this.expectedChildIdentifier?.toString() ?? 'undefined';
@@ -519,12 +522,10 @@ class JessieExpressionVisitor extends NodeVisitor<
         this.logger,
         {
           ...this.stack,
-          input: {
-            ...(this.inputParameters ?? {}),
-          },
-          parameters: {
-            ...(this.integrationParameters ?? {}),
-          },
+          ...fromEntriesOptional(
+            ['input', this.inputParameters],
+            ['parameters', this.integrationParameters]
+          )
         }
       );
 
@@ -702,7 +703,7 @@ class CallVisitor extends NodeVisitor<
 type HttpRequest = {
   contentType?: string;
   contentLanguage?: string;
-  headers?: Variables;
+  headers?: NonPrimitive;
   queryParameters?: NonPrimitive;
   body?: Variables;
   security: HttpSecurityRequirement[];
@@ -771,7 +772,13 @@ class HttpCallStatementVisitor extends NodeVisitor<HttpCallStatementNode> {
           accept,
           baseUrl: serviceUrl,
           queryParameters: request?.queryParameters,
-          pathParameters: { ...this.stack, input: this.inputParameters, parameters: this.integrationParameters },
+          pathParameters: {
+            ...this.stack,
+            ...fromEntriesOptional(
+              ['input', this.inputParameters],
+              ['parameters', this.integrationParameters]
+            )
+          },
           body: request?.body,
           securityRequirements: request?.security,
           securityConfiguration: this.securityConfiguration,
@@ -1173,7 +1180,7 @@ export class MapInterpreter<TInput extends NonPrimitive | undefined> {
     for (const value of Object.values(input)) {
       if (isInitializable(value)) {
         await value.initialize();
-      } else if (value !== undefined && value !== null && isNonPrimitive(value)) {
+      } else if (isNonPrimitive(value)) {
         await MapInterpreter.initializeInput(value);
       }
     }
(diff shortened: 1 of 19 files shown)


freaz:
Before diving to implementation, I want quickly share opinion:

1. relaxing headers and query parameters values: I think it is good idea, I would keep in spec though it must be string and do casting in OneSDK only.
2. What I completely forgot are defaults from `super.json`, but I assume to the interpreters perform is passed already merged value of input and integration parameters with defaults from `super.json`. I would take advantage of control under Comlinks in station and before releasing, test it against to see if it even happened.

src/node/fetch/fetch.node.ts
```diff
@@ -148,26 +148,21 @@ export class NodeFetch implements IFetch, Interceptable, AuthCache {
...
+    if (parameters === undefined || Object.keys(parameters).length === 0) {
+      return '';
+    }
 
-      return '?' + new URLSearchParams(definedParameters).toString();
+    const params = new URLSearchParams();
+    for (const [key, param] of Object.entries(parameters)) {
+      if (typeof param === 'string') {
+        params.append(key, param);
+      } else {
```
freaz, line 160: Hopefully we can rely on types.

freaz (commented):
Besides the issue with duplicate headers, there are just some suggestions, which can be added in follow ups.

src/core/interpreter/map-interpreter.ts
```diff
@@ -702,7 +703,7 @@ class CallVisitor extends NodeVisitor<
 type HttpRequest = {
   contentType?: string;
   contentLanguage?: string;
-  headers?: Variables;
+  headers?: NonPrimitive;
```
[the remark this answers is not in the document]
TheEdward162: This is still in interpreter and there we can only promise NonPrimitive as the visitor doesn't check the types. We later check the types in the http module (in `variablesToHttpMap`).

src/node/fetch/fetch.node.ts
TheEdward162: From how it is structured now this MultiMap comes from our variablesToHttpMap. I would also argue that if we can't trust the types at runtime we might as well not have them and do typeof checks everywhere, but I'd rather not do that 😅

freaz: Well I am concerned, because of this line https://github.com/superfaceai/one-sdk-js/blob/3d95f6169d4ad669f5f297173bb953b61ef7b622/src/core/interpreter/map-interpreter.ts#L899

src/core/interpreter/map-interpreter.ts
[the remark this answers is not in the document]
freaz: Ok, now I understand. thank you.

src/node/fetch/fetch.node.ts
TheEdward162: It still goes through `variablesToHttpMap` (and `variablesToString` before these changes) before the request is dispatched to the fetch implementation, which ensures its type.

The cast is there because right now there is no way to infer the return type of `yield` based on input to `yield` - i.e.:
```
function* gen(): Generator<1 | 2, undefined, 'one' | 'two'> {
	const foo = yield 1;
	const bar = yield 2;

	return undefined;
}

const it = gen();
let next = it.next();
while (next.done !== true) {
	let stringified: 'one' | 'two';
	if (next.value === 1) {
		stringified = 'one';
	} else {
		stringified = 'two';
	}
	
	next = it.next(stringified);
}
```

There is no way to tell typescript that `foo: 'one'` and `bar: 'two'` even when we know this to be true. Same issue with the typecast - we are exploring `ObjectLiteralNode` which always returns `NonPrimitive`. I left the TODOs in because I wasn't sure whether we should typecheck, but thinking about it now I think it's not necessary. They should instead be SAFETY comments explaining why the cast is safe. This is an implementational detail and it should be tested, but it's not a reason to add a typecheck in a wholly separate class which may very well be in another package.

freaz: Ok, thank you for explanation.

freaz (approved):

PR #324, opened 2023-01-17
Multi-valued response headers as multimap

TheEdward162:
This applies on top of #323. It was supposed to fix multi-valued response headers, but it seems we are blocked by node-fetch and upgrading it to 3.x version isn't trivial, so for now only these changes.

I'm setting the base to #323 for the diff to be meaningful, but it should be changed to `dev` later, I suppose.

### DIFF
--- a/src/node/fetch/fetch.node.test.ts
+++ b/src/node/fetch/fetch.node.test.ts
@@ -8,19 +8,15 @@ import { MockTimers } from '../../mock';
 import { NodeTimers } from '../timers';
 import { NodeFetch } from './fetch.node';
 
-// Maybe this can be done better?
-// The issue here is that we need to mock the default export (fetch) while
-// preserving everything as as actual implementation.
 jest.mock('node-fetch', () => {
-  const actual = jest.requireActual('node-fetch');
+  // eslint-disable-next-line @typescript-eslint/consistent-type-imports
+  const actual = jest.requireActual<typeof import('node-fetch')>('node-fetch');
 
-  const base: any = jest.fn();
-  for (const [key, value] of Object.entries(actual)) {
-    base[key] = value;
-  }
-  base['default'] = base;
-
-  return base as unknown;
+  return {
+    __esModule: true,
+    ...actual,
+    default: jest.fn(),
+  };
 });
 
 const mockServer = getLocal();
--- a/src/node/fetch/fetch.node.test.ts
+++ b/src/node/fetch/fetch.node.test.ts
@@ -11,9 +11,9 @@ import { NodeFetch } from './fetch.node';
 // Maybe this can be done better?
 // The issue here is that we need to mock the default export (fetch) while
 // preserving everything as as actual implementation.
-jest.mock('node-fetch', ()  => {
+jest.mock('node-fetch', () => {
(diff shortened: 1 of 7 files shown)


src/node/fetch/fetch.node.test.ts
```diff
@@ -1,19 +1,31 @@
...
+  const actual = jest.requireActual('node-fetch');
+  
+  const base: any = jest.fn();
+  for (const [key, value] of Object.entries(actual)) {
+    base[key] = value;
+  }
+  base['default'] = base;
+
+  return base as unknown;
+});
```
freaz, line 24: Not sure if it is better, but I usually do it this way:

```suggestion
jest.mock('node-fetch', () => {
  // eslint-disable-next-line @typescript-eslint/consistent-type-imports
  const actual = jest.requireActual<typeof import('node-fetch')>('node-fetch');

  return {
    __esModule: true,
    ...actual,
    default: jest.fn(),
  };
});
```

freaz (commented):
Thanks for all changes. All headers handling look much cleaner now.

It is just a pity we can't make it RFC compliant, but it is what it is.

I suggested one change, but also have ready commit with skipping and commenting out problematic tests with explanation in commit. If it is OK with you, I will push it to your branch. 🙏

src/node/fetch/fetch.node.test.ts
TheEdward162: As long as that works 😅 I wasn't sure at all how to mock this properly so if you usually do it this way let's stick with that.

src/node/fetch/fetch.node.test.ts
```diff
@@ -390,34 +374,44 @@ describe('NodeFetch', () => {
...
-  describe('prepareHeadersInit', () => {
-    let nodeFetch: NodeFetch;
 
-    type OverrideNodeFetch = {
-      prepareHeadersInit: (data: any) => any | undefined;
-    };
+    // this test works under the assumption that node-fetch returns multi-valued headers as arrays
+    // this is not true for node-fetch 2.x
+    // eslint-disable-next-line jest/no-disabled-tests
+    it.skip('should correctly send and receive multi-valued headers', async () => {
```
janhalama, line 381: question: I tried to run the test and it actually passed, even though the node-fetch version is node-fetch 2.6.7. Why is the test skipped?

janhalama (commented):
I went through the code, it looks good to me. I have one question. Be aware that the base branch is set to `feat/none_type`.

src/node/fetch/fetch.node.test.ts
freaz: Yeah, it is passing because we are mocking `fetch` and mocked function returns header values as an array as it should. I just skipped it, but I will also change it to use fetch instance, so it is not confusing.

janhalama: Thanks for explanation. The mock was actually behaving according to node-fetch v3.

freaz: Exactly mock was working as it would be result from node-fetch 3.0 instead of actually used 2.0.

PR #339, opened 2023-02-27
Normalize superjson preserve ast field

freaz:
## Description

Follow up to #336 

Normalize function didn't take in account `ast` field and created normalized document without it.

## Types of changes

## Checklist:

### DIFF
--- a/src/schema-tools/superjson/normalize.ts
+++ b/src/schema-tools/superjson/normalize.ts
@@ -28,6 +28,7 @@ import type { IEnvironment, ILogger } from '../../interfaces';
 import {
   castToNonPrimitive,
   clone,
+  isSettingsWithAst,
   mergeVariables,
   resolveEnvRecord,
   UnexpectedError,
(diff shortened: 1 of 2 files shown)


src/schema-tools/superjson/normalize.ts
```diff
@@ -332,9 +345,27 @@ export function normalizeProviderSettings(
...
+      file: providerEntry.file,
+      security: providerEntry.security ?? [],
+      parameters: providerEntry.parameters ?? {},
+    };
+  }
+
   if (environment !== undefined) {
     return {
-      file: providerEntry.file,
+      ...normalizedSettings,
```
TheEdward162, line 368: question: Wouldn't it be better to assign to `security: normalizedSettings.security.map(...resolveEnvRecord...)`? That way even if someone changes `security` and `parameters` in the ifs above it will propagate into the result object.

TheEdward162 (approved):

src/schema-tools/superjson/normalize.ts
freaz: 🤦‍♂️definitely yes. Thanks

